PR #7, opened 2019-10-19
WIP: Add konduit CLI

wmeddie:
Add a cli module that builds to a native executable.


konduit-cli/Readme.md
```diff
@@ -0,0 +1,16 @@
+# CLI for Konduit
+
+using the Konduit CLI you can easily create high performance model servers
+from simple python scripts with arbitrarily complex pre/post processing.
```
maxpumperla: is this for python only? if so, it might make sense to build it directly with `click` in Python, to attract the community to participate in fixes or extensions.

wmeddie: Not only python. It’ll work for anything that we can configure Konduit with. I wanted to make this a native CLI app so that it could be used without relying on the user having to have anything installed. I think we can still implement a plugin system but I think most things would be more useful in the server side... need to think about that more.

konduit-cli/pom.xml
```diff
@@ -0,0 +1,129 @@
...
+        <dependency>
+            <groupId>info.picocli</groupId>
+            <artifactId>picocli</artifactId>
+            <version>${picocli.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>${junit.version}</version>
```
agibsonccc: Parent module already has this dependency managed as test scope: https://github.com/KonduitAI/konduit-serving/blob/4f5e01525160719e49057a4b4425a03b6688f325/pom.xml#L323

konduit-cli/src/main/java/ai/konduit/serving/KonduitCli.java
```diff
@@ -0,0 +1,22 @@
+package ai.konduit.serving;
+
+import ai.konduit.serving.ai.konduit.serving.commands.PingCommand;
+import picocli.CommandLine;
+import picocli.CommandLine.Command;
+import picocli.CommandLine.Option;
+
+@Command(name = "konduit", version = "1.0.0", subcommands = {
```
agibsonccc: Is there a way to tie this to project.version using maven metadata?

konduit-cli/src/test/java/ai/konduit/serving/KonduitCliTest.java
```diff
@@ -0,0 +1,35 @@
...
+    @Test
+    public void shouldPingCorrectly() {
+        int exitCode = new CommandLine(new KonduitCli()).execute("ping");
+
+        assertEquals(0, exitCode);
+    }
+
+    @Test
+    public void shouldShowHelp() {
+        KonduitCli cli = new KonduitCli();
```
agibsonccc: How do we plan on executing this from python? Jnius?

konduit-cli/pom.xml
```diff
@@ -0,0 +1,129 @@
...
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
```
agibsonccc: Wondering if maybe we can consolidate the compiler plugin definitions under pluginManagement instead? Then you can just add the necessary over rides rather than readding the version etc each time

maxpumperla:
@wmeddie given that the "other" CLI is already built and in use (plus it's about 10% of the code, easier to maintain), I wonder if it makes sense to keep this open for now.

The main benefit I see from this PR, as discussed in the RFC, is that it promises to deploy pipelines to various (remote) server configurations. once we have the backend functionality for that, it will be easy to expose it to any CLI, either this one or the other.

PR #24, opened 2019-10-24
Basic Python example for TensorFlow serving

philip-khor:
Modified version of test_bert_serving.py for a user-facing example of the Python SDK.

### DIFF
--- a/python/examples/01_tensorflow-basic.ipynb
+++ b/python/examples/01_tensorflow-basic.ipynb
@@ -40,7 +40,21 @@
     "2. One or more machine learning models\n",
     "3. Transforming the output in a way that can be understood by humans, such as labels in a classification example.\n",
     "\n",
-    "If deploying your model does not require pre- nor post-processing, only one pipeline step - a machine learning model - is required. This configuration is defined using a single `ModelPipelineStep`. "
+    "If deploying your model does not require pre- nor post-processing, only one pipeline step - a machine learning model - is required. This configuration is defined using a single `ModelPipelineStep`. \n",
+    "\n",
+    "Before running this notebook, you should run the `build_jar.py` script and copied the JAR to this folder. \n",
+    "\n",
+    "Start by downloading the model weights to the `data` folder. "
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "!curl https://deeplearning4jblob.blob.core.windows.net/testresources/bert_mrpc_frozen_v1.zip --output ../data/bert.zip\n",
+    "!unzip ../data/bert.zip "
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


python/examples/01_tensorflow-basic.ipynb
```diff
@@ -0,0 +1,290 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Configuring the server\n",
+    "\n",
+    "Now we configure the server. Here, specify the following:\n",
+    "- `input_data_type`: select a random port.\n",
```
ShamsUlAzeem: Typo here. `input_data_type` -> `http_port`

python/examples/01_tensorflow-basic.ipynb
```diff
@@ -0,0 +1,290 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "!curl https://deeplearning4jblob.blob.core.windows.net/testresources/bert_mrpc_frozen_v1.zip --output ../data/bert.zip\n",
+    "!unzip ../data/bert.zip "
```
ShamsUlAzeem: Can you mention OS specific commands so these can run inside Windows, MacOS and Linux. If they need to install anything then you can mention that as well. A good way to handle some items is to write the download and extration script in python

python/examples/01_tensorflow-basic.ipynb
```diff
@@ -0,0 +1,290 @@
...
+   "metadata": {},
+   "source": [
+    "### Configuring the client \n",
+    "\n",
+    "To configure the client, create a Client object with the following arguments: \n",
+    "- `input_names`: names of the input data\n",
+    "- `output_names`: names of the output data\n",
+    "- `input_type`: data type passed to the server for inference\n",
+    "- `endpoint_output_type`: data type returned by the server endpoint \n",
+    "- `return_output_type`: data type to be returned to the client\n",
```
ShamsUlAzeem: Also, mention that this can be used to convert the output returned from the server to the client into a different format. For example, NUMPY to JSON etc etc.

python/examples/01_tensorflow-basic.ipynb
```diff
@@ -0,0 +1,290 @@
...
+   "metadata": {},
+   "source": [
+    "## Running the server "
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We load some sample data from NumPy files. Note that these are NumPy arrays, each with shape (4, 128): "
```
ShamsUlAzeem: "We will load" instead of "We load"

python/examples/01_tensorflow-basic.ipynb
```diff
@@ -0,0 +1,290 @@
...
+    "    'IteratorGetNext:1': np.load('../data/input-1.npy'),\n",
+    "    'IteratorGetNext:4': np.load('../data/input-4.npy')\n",
+    "}"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We start the server and wait 60 seconds for the server to start before the client requests the server for a prediction using the `data_input`. "
```
ShamsUlAzeem: "We will start"

python/examples/01_tensorflow-basic.ipynb
philip-khor: Replaced as follows: 

```python 


from urllib.request import urlretrieve 
from zipfile import ZipFile
urlretrieve("https://deeplearning4jblob.blob.core.windows.net/testresources/bert_mrpc_frozen_v1.zip", "../data/bert.zip")
with ZipFile('../data/bert.zip', 'r') as zipObj:
    zipObj.extractall()
```

python/examples/01_tensorflow-basic.ipynb
philip-khor: Reworded to be "Load some sample data ..." to be consistent with the rest of the document.

python/examples/01_tensorflow-basic.ipynb
philip-khor: Replaced with "Start the server ..." to be consistent with the rest of the doc.

philip-khor:
> 
> 
> Oh, one last thing here. Can you execute each cell so that the output is shown?

done!

PR #30, opened 2019-10-29
Fixed naming to KonduitServingMain

chiaweilim:
Found that the mainclass in pom.xml did not aligned with the class name when building the uberjar.

Changed the naming from KonduitServerMain to KonduitServingMain


ShamsUlAzeem:
@chiaweilim great catch. Btw do you want to include the notebooks in the notes folder in the repo as well?

ShamsUlAzeem:
Merging this

PR #36, opened 2019-11-04
API changes, Examples and Python Docker image work

ShamsUlAzeem:
Just an initial iteration on the API changes, relevant examples and python pipeline step for docker image


ShamsUlAzeem:
The conflicts are due to refactored classes for orchestration. Will solve them at the end of finishing this PR.

konduit-serving-api/src/main/java/ai/konduit/serving/pipeline/PipelineStep.java
```diff
@@ -135,7 +135,7 @@ public PipelineStep input(String[] columnNames, SchemaType[] types) throws Excep
      * @return this pipeline step
      * @throws Exception key error
      */
-    public PipelineStep setInput(String inputName, String[] columnNames, SchemaType[] types)
+    public PipelineStep input(String inputName, String[] columnNames, SchemaType[] types)
```
maxpumperla: I changed this because `input` is a protected keyword in python and I wanted to keep things consistent with `setInput` and `set_input` on the Java and Python sides respectively.

konduit-serving-api/src/main/java/ai/konduit/serving/pipeline/PythonPipelineStep.java
```diff
@@ -93,8 +93,8 @@ public PythonPipelineStep(PythonConfig pythonConfig, String[] inputColumnNames,
     public PythonPipelineStep(Schema inputSchema, Schema outputSchema,
                               PythonConfig pythonConfig) throws Exception {
         String defaultName = "default";
-        this.setInput(defaultName, inputSchema);
-        this.setOutput(defaultName, outputSchema);
```
maxpumperla: same here, you decide. we can leave it like this.

docker/konduit-python/start-server.py
```diff
@@ -71,37 +71,39 @@ def schema_type(python_type):
...
+                    output_names=[default],
                     input_type='NUMPY',
                     endpoint_output_type='NUMPY',
-                    url='http://localhost:' + str(port))
+                    url='http://localhost:' + str(port), timeout=2000000)
 
     data_input = {
-        'x': np.ones([2, 2]),
-        'y': np.ones([2, 2])
+        'default:x': np.ones([2, 2]),
```
maxpumperla: yeah, we definitely need to fix this behavior in the next PR. it's very confusing.

konduit-serving-api/src/main/java/ai/konduit/serving/pipeline/PipelineStep.java
ShamsUlAzeem: Sure that makes sense. I updated it to the set* signature https://github.com/KonduitAI/konduit-serving/pull/36/commits/c9232f657779764d89447dd8397da154f11991d1

konduit-serving-api/src/main/java/ai/konduit/serving/pipeline/PythonPipelineStep.java
ShamsUlAzeem: c9232f6

docker/konduit-python/start-server.py
ShamsUlAzeem: I need to look at the BatchInputParser class a bit before we can merge

docker/konduit-python/run.sh
```diff
@@ -0,0 +1,9 @@
+echo "Installing required packages..."
+pip install --no-cache-dir -r requirements.txt
+echo "Python packages installed successfully."
+
+echo "-------------------------------------------------------------------"
+
+echo "Starting konduit server..."
```
agibsonccc: I'd allow a potential over ride in the docker container. If you want to pre package a python script execution experience, have a default script in there and allow a volume mounting an over rideable script in to the container. Maybe add a docker compose example in there along side that.

docker/konduit-python/start-server.py
```diff
@@ -0,0 +1,115 @@
...
+    # ----------------------------------------------------------------------------------------
+
+    port = random.randint(1000, 65535)
+
+    serving_config = ServingConfig(http_port=port,
+                                   input_data_type=rest_input_type,
+                                   output_data_type=rest_output_type,
+                                   log_timings=True)
+
+    pythonConfig = {
```
agibsonccc: Just of note is you can use a to json utility present right in konduit-serving python: https://github.com/KonduitAI/konduit-serving/blob/master/python/konduit/json_utils.py

konduit-serving-api/pom.xml
```diff
@@ -53,6 +53,11 @@
     </profiles>
 
     <dependencies>
+        <dependency>
```
agibsonccc: Why use this vs the vertx json objects or nd4j jackson? This seems un needed.

konduit-serving-api/src/main/java/ai/konduit/serving/pipeline/ModelPipelineStep.java
```diff
@@ -23,26 +23,67 @@
...
 
     private NormalizationConfig normalizationConfig;
 
+    public ModelPipelineStep() {}
+
+    public ModelPipelineStep(ModelConfig modelConfig) {
+        this.modelConfig = modelConfig;
+    }
+
+    public ModelPipelineStep setInput(String[] columnNames, SchemaType[] types) throws Exception {
```
agibsonccc: Kinda wondering if we could make use of generics here and clean this up a bit.

docker/konduit-python/run.sh
ShamsUlAzeem: @agibsonccc yes at the end the user just need to override the script with a volume using `-v`

PR #40, opened 2019-11-06
basic jar building API

maxpumperla:
This works once released: 

```
pip install konduit
konduit-python init --os <your_os>
```


maxpumperla:
p.s. @wmeddie I called the cli tool `konduit-python` so as to not interfere with your tool. but we could rename it at any point.

PR #43, opened 2019-11-12
Remove pubsub configuration

### LINKED ISSUE #41: Remove old PubSubConfig
The pubsub functionality with http isn't in there anymore. We can delete the tests and reintroduce an a more explicit pubsub abstraction instead (using an abstraction over grcp, kafka, and http).

agibsonccc:
Closes #41 , our serialization already is robust to removed properties so there shouldn't need to be too much that happens here.


python/konduit/base_inference.py
```diff
@@ -46,61 +46,6 @@ def as_dict(self):
...
-        if self.__http_method is not None:
-            d['httpMethod'] = self.__http_method.as_dict() if hasattr(
-                self.__http_method, 'as_dict') else self.__http_method
-        if self.__pubsub_url is not None:
-            d['pubsubUrl'] = self.__pubsub_url.as_dict() if hasattr(
-                self.__pubsub_url, 'as_dict') else self.__pubsub_url
-        if self.__content_type is not None:
-            d['contentType'] = self.__content_type.as_dict() if hasattr(
-                self.__content_type, 'as_dict') else self.__content_type
-        return d
```
maxpumperla: you're missing the `ServingConfig`, which still has a pub sub config

agibsonccc:
@maxpumperla go ahead and review

maxpumperla:
thanks

PR #47, opened 2019-11-13
Remove graph mapper getInstance()

### LINKED ISSUE #46: Due to snapshots, TFGraphMapper in samediff still uses old getInstance()
We need to update this to reflect the new samediff behavior.

agibsonccc:
Closes #46 

@sshepel


maxpumperla:
see, this one got merged automatically since it was literally the same commit hash

PR #51, opened 2019-11-13
python step directly from config

maxpumperla:
the Python side of things will go in a follow-up PR


ShamsUlAzeem:
@maxpumperla some of the `target` folder classes creeped inside. Other than that it's great

PR #53, opened 2019-11-13
next round of python improvements

maxpumperla:


python/tests/test_python_serving.py
```diff
@@ -10,35 +10,19 @@
...
-    serving_config = ServingConfig(http_port=port,
-                                   input_data_type='NUMPY',
-                                   output_data_type='NUMPY',
-                                   log_timings=True)
+    serving_config = ServingConfig(http_port=port)
 
     python_config = PythonConfig(
         python_code='first += 2',
         python_inputs={'first': 'NDARRAY'},
         python_outputs={'first': 'NDARRAY'},
```
ShamsUlAzeem: @maxpumperla can we create enums here too? `VariableTypes.NDARRAY` or maybe `VariableTypes.Python.NDARRAY` and `VariableTypes.SchemaType.NDArray`

ShamsUlAzeem:
@maxpumperla adam has some changes request that are comment on the pull request here: https://github.com/KonduitAI/konduit-serving/pull/57. I've mentioned your name where your attention is needed.

maxpumperla:
@ShamsUlAzeem thanks for the feedback. I've noted your idea about enums. Let's come back to that at a later point, for now I think the way it is is easier for Python folks.

PR #64, opened 2019-11-15
Removed unnecessary target files

ShamsUlAzeem:
Deleting crept in target files


ShamsUlAzeem:
@agibsonccc the target folders are already inside .gitignore. This file might have got in from the reupload somehow. Did a little bit of clean up though.

ShamsUlAzeem:
Ran tests and nothing comes in from the target folder now

PR #71, opened 2019-11-16
Purge multipartuploadhandler

### LINKED ISSUE #72: Remove old BatchInputParser interface
BatchInputParser had old ideas attached to it that went un used. We should remove the upload handler listener.

agibsonccc:
Removes multi part handler since it's an unused interface.
Closes #72


maxpumperla:
@agibsonccc looking at your other PR, it seems like this one is redundant. correct?

agibsonccc:
@maxpumperla no this was meant to be a self contained change. There are further changes I'm doing on that PR.

PR #77, opened 2019-11-19
Update detect_image.py

philip-khor:
Pretrained model file for PyTorch example renamed in commit Linzaer/Ultra-Light-Fast-Generic-Face-Detector-1MB@0f9ca4a


maxpumperla:
@philip-khor good catch

PR #83, opened 2019-11-20
doctest integration

maxpumperla:
@ShamsUlAzeem @philip-khor @chiaweilim as discussed this morning, we'd like to have some basic API checks for our docs. This PR establishes that baseline.

In particular, the new test `test_docs.py` will go through all markdown files in a specified folder (for now `docs`) and extract all Python executable code and runs it. This test will fail when examples from docs are not executable. In particular, pull requests that break the docs will not pass CI. That is good news and once we establish a good progress for doing the bidirectional sync with gitbook for our docs, we should be in a much better state.


agibsonccc:
@maxpumperla one thing I'd be interested in is a jnius based exec of java based code blocks if we could do the conversion.

ShamsUlAzeem:
Oh, that's a good point. We can implement a similar code block using ````java`. But yeah it would need conversions from java based code to pyjnius python equivalent code.

maxpumperla:
@agibsonccc yeah, that should be doable as well. we could generate some boilerplate and paste the code bits from docs into a junit test that is then simply called by pyjnius (in fact, a lot of the unit tests from pyjnius are run against a predefined junit test).

I'll provide a basic example for the crew to pick up

maxpumperla:
@agibsonccc @ShamsUlAzeem works for Java now as well. we might want to do something like this for OSS docs as well.

python/tests/test_docs.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+def write_java_files(code_blocks):
+    make_folders()
+    for i, block in enumerate(code_blocks):
+        lines = block.split('\n')
+        import_lines = [l for l in lines if l.startswith('import ')]
+        code_lines = [l for l in lines if not l.startswith('import ')]
+
+        code = "package ai.konduit.serving;\n\n" + '\n'.join(import_lines) \
```
agibsonccc: You could just wrap this in a main method and run it like that maybe? Also - what about maybe having a java file you load as a template we bundle instead? I know it might seem annoying but it would make this a bit more maintainable than stringly typed code :) That's what I did to the python executioner.

python/tests/test_docs.py
```diff
@@ -0,0 +1,121 @@
+from konduit import *
+
+
+def markdown_blocks(file_path, language='python', safe=True):
+    code_blocks = []
+    code_regex = r'^```.*'
+    code_block_open_re = r'^```(`*)({0}){1}$'.format(language, '' if safe else '?')
```
agibsonccc: Kinda wondering if there's a more robust markdown parser we could use in place of this? Do you see any edge cases coming from this?

python/tests/test_docs.py
maxpumperla: @agibsonccc sure, let me think about a "main" runner (right now I add the class to the konduit.jar, so that you only need to do that once. unit tests run really fast like this).

well, about the template. you see that it's literally 5 lines of boiler plate, the rest comes from code snippets in docs. and those snippets _are_ strings, no matter how we build it in the end. also, keep in mind that I build one java class per code snippet to ensure they're not entangled and each snippet runs standalone as expected. not sure how I would to that with templates (I don't know how many I need, for starters)

python/tests/test_docs.py
maxpumperla: good point, let me investigate. edge cases can only arrive from human error, we're literally just checking what comes between opening and closing triple backticks (plus a safeguard for the language)

maxpumperla:
@agibsonccc @ShamsUlAzeem we can now also test standalone "main method" java classes, if that should become necessary for our docs efforts.

ShamsUlAzeem:
Awesome!!!

PR #93, opened 2019-11-24
Merges mem map verticle with inference verticle

### LINKED ISSUE #70: Merge MemMapVerticle in to InferenceVerticle
@ShamsUlAzeem @maxpumperla  @wmeddie  https://github.com/KonduitAI/konduit-serving/blob/4f5e01525160719e49057a4b4425a03b6688f325/konduit-serving-core/src/main/java/ai/konduit/serving/verticles/nd4j/memmap/MemMapVerticle.java#L69

What do we want MemMapVerticle doing? Do we want to merge this in to InferenceVerticle as a configuration instead?

Do we want more than one verticle? Collapsing them seems like it will make things easier.
I know for example, we could also just make mem map a pipeline step instead.

### LINKED ISSUE #80: PMML tests missing
Somewhere along the way of moving to the separate `konduit-serving-pmml` package we literally lost all tests and examples. the test resources are still there, we just need to pick them up again.

agibsonccc:
Closes #70  and also fixes #80


konduit-serving-api/src/main/java/ai/konduit/serving/util/SchemaTypeUtils.java
```diff
@@ -260,6 +260,7 @@ public static Schema toSchema(SchemaType[] types, List<String> names) {
             List<INDArray> accum = new ArrayList<>();
             //for each record
             for(Record record : records) {
+                Writable curr = record.getRecord().get(i);
```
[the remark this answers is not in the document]
ShamsUlAzeem: Ah, I think this PR is still in progress.

python/konduit/base_inference.py
```diff
@@ -60,89 +64,85 @@ class SavedModelConfig(object):
...
+    
     def _get_saved_model_input_order(self):
         return self.__saved_model_input_order
-
     def _set_saved_model_input_order(self, value):
-        if not isinstance(value, list) and not isinstance(value, ListWrapper):
+        if not isinstance(value, list) and not isinstance(value,ListWrapper):
             raise TypeError("savedModelInputOrder must be list")
         if not all(isinstance(i, str) for i in value):
             raise TypeError("savedModelInputOrder list valeus must be str")
```
ShamsUlAzeem: There's this `valeus` typo at several places.

agibsonccc: @maxpumperla let's use this as an excuse to update the code gen to make sure we have proper output values. Thanks @ShamsUlAzeem

PR #92, opened 2019-11-23
python fixes

farizrahman4u:

PR #96, opened 2019-11-25
python fixes + make setuprun optional

farizrahman4u:


konduit-serving-python/src/main/java/ai/konduit/serving/executioner/PythonExecutioner.java
```diff
@@ -607,13 +607,13 @@ public static void execWithSetupAndRun(String code) {
         acquireGIL();
         _exec(code);
         if (_hasGlobalVariable("setup") && _hasGlobalVariable("run")){
-            log.debug("setup() and run() methods found.");
+            log.info("setup() and run() methods found.");
```
agibsonccc: Mind adding comments around context in the various run modes we're doing here?

konduit-serving-python/src/main/java/ai/konduit/serving/executioner/PythonExecutioner.java
```diff
@@ -788,6 +807,10 @@ public static PythonVariables execAndReturnAllVariables(String code, PythonVaria
         String inputCode = inputCode(pyInputs);
         return execAndReturnAllVariables(inputCode + code);
     }
+    public static PythonVariables execWithSetupRunAndReturnAllVariables(String code, PythonVariables pyInputs) throws Exception {
```
agibsonccc: Given all the different static methods here, I wonder if we should just have a builder object and flags instead? @maxpumperla ?

konduit-serving-python/src/main/java/ai/konduit/serving/executioner/PythonExecutioner.java
farizrahman4u: added in #92

konduit-serving-python/src/main/java/ai/konduit/serving/executioner/PythonExecutioner.java
maxpumperla: @agibsonccc in Python this will translate into a single method with default args. In the end users will only use this through `PythonTransform`, i.e. eventually in `PythonStep`. So I wouldn't worry about this.

agibsonccc:
cc @farizrahman4u

agibsonccc:
@farizrahman4u  let's go ahead and do what @maxpumperla suggested in a separate PR.

PR #98, opened 2019-11-27
Python Client and Server validation

ShamsUlAzeem:
Since we were using a sleep function for giving the server a buffer time for loading, the time to load depending on each configuration was ambiguous. This PR implements simple request retries for checking server health using the `/healthcheck` endpoint. We can add more functionality to the health checks but for now it returns status code of `204` with no content if the server is up and running. You can increase the `start_timeout` while creating the `Server` object if you think the network will take more than 2 minutes to start. That variable specified how long the internal mechanism should retry health checks til reporting a service status.

Other than that this PR also validates the input and output names based on what the server is configured for. This way we won't have to specify some of the input and output names while creating a client and it will be loaded from the server through `/config` or `/config/pretty` endpoints.


ShamsUlAzeem:
@agibsonccc do I have to add the same endpoints to `MemMapRouteDefiner`?

python/konduit/server.py
```diff
@@ -22,7 +22,8 @@ def stop_server_by_pid(pid):
 class Server(object):
     def __init__(self, inference_config=None, serving_config=None, steps=None,
                  extra_start_args='-Xmx8g', config_path='config.json',
-                 jar_path='konduit.jar'):
+                 jar_path='konduit.jar', pid_file_path='konduit-serving.pid',
```
maxpumperla: is `pid_file_path` really an argument to this Python object? We create this file server-side, but at least in the current version it's only stored in Python, i.e. if I change the name of the file here, the backend won't be affected.

python/konduit/server.py
```diff
@@ -75,8 +91,32 @@ def start(self, sleep=None):
...
+        tries = int(self.start_timeout/request_timeout + 1)
+
+        started = False
+
+        for i in range(tries):
+            # This url returns status 204 with no content when the server is up.
+            try:
+                print("Checking server integrity. Tries: {} of {}".format(i + 1, tries))
+
+                r = requests.get("http://localhost:{}/healthcheck".format(self.config.serving_config.http_port),
```
maxpumperla: I guess right now it doesn't make much of a difference, but shall we really hard-code `localhost` here? we usually only do this in tests.

This points to a deeper issue, I guess. Where does a `Server` actually get deployed and how can we control this conveniently? Pretty sure we don't always want to do `localhost`. However, right now `Server` is just created like this, while `Client` actually requires a URL to connect. If anything, I'd assume the server determines the url and the client connects against the server according to the server's config.

This is out of scope for this PR, I think, but still something we should look into.

python/konduit/server.py
```diff
@@ -75,8 +91,32 @@ def start(self, sleep=None):
...
+                                 timeout=request_timeout)
+                if r.status_code == 204:
+                    started = True
+                    break
+            except Exception as ex:
+                logging.debug("{}\nChecking server integrity again...".format(str(ex)))
+
+        if started:
+            print("Server has started successfully.")
```
maxpumperla: consider using `logging.info` in favor of print

python/konduit/client.py
```diff
@@ -10,7 +10,31 @@
 
 class Client(object):
     def __init__(self, timeout=60, input_type='NUMPY', endpoint_output_type='NUMPY',
-                 return_output_type=None, input_names=['default'], output_names=['default'], url=None):
+                 return_output_type=None, url=None):
+
+        try:
+            r = requests.get("{}/healthcheck".format(url))
+            if r.status_code != 204:
```
maxpumperla: I was thinking, maybe we can make this `try` block a utility function to be used in other places. because if we don't have that, how would users verify that the server is running?

python/konduit/client.py
```diff
@@ -10,7 +10,31 @@
 
 class Client(object):
     def __init__(self, timeout=60, input_type='NUMPY', endpoint_output_type='NUMPY',
-                 return_output_type=None, input_names=['default'], output_names=['default'], url=None):
+                 return_output_type=None, url=None):
+
+        try:
+            r = requests.get("{}/healthcheck".format(url))
```
maxpumperla: Also not really related to this PR, but it's pretty confusing that one needs to specify just a `port` in the serving config and then remember to put it into the `url` when creating the client. maybe we can have a basic to assert that `url.split(os.pathsep)[1]` can be cast to an int etc.

python/tests/test_client.py
```diff
@@ -5,7 +5,7 @@
 
 def test_multipart_regex():
 
-    client = Client(input_names=['partname'], output_names=['nobody_cares'])
+    client = Client()
```
maxpumperla: that breaks the test. you can remove the output names, but not the input

python/konduit/server.py
ShamsUlAzeem: @maxpumperla how about we have something like. `server.get_client()`. We don't really have any workflow right now in which we're creating the server remotely since it depends on a local `konduit.jar` file to run so I've `localhost` in there for now.

python/konduit/server.py
ShamsUlAzeem: how do we change the log level inside the python `logging` module
I added the print statements because it wasn't printing it on the notebooks except for the `error` level

python/konduit/server.py
ShamsUlAzeem: I think the Server class would always be at localhost , no? Since we're always creating it from a local `konduit.jar` file.

python/konduit/server.py
ShamsUlAzeem: Only thing is that I need to change the location of that pid file while starting that server
I need to override this somehow through the python code
https://github.com/KonduitAI/konduit-serving/blob/3d726a1c377c7753678da012cee2aff158695b0a/konduit-serving-core/src/main/java/ai/konduit/serving/configprovider/KonduitServingNodeConfigurer.java#L120
or maybe get that as info from the server

python/konduit/client.py
ShamsUlAzeem: Ah that's a good idea for sure.

python/konduit/client.py
ShamsUlAzeem: Sure. That can be added

python/tests/test_client.py
ShamsUlAzeem: Hmm. Maybe then I'll just leave the input names and output names arguments in there and if they're None then I'll try to infer them from the server

python/konduit/server.py
maxpumperla: server.get_client() is essentially what I did in the other PR, but with a function, i.e. client_from_server(server). we can move that. I like this idea of yours.

python/konduit/server.py
ShamsUlAzeem: oh `logging.getLogger().setLevel(logging.DEBUG)`
I think I'll leave in the print statements because then it will always give out some extra info that I don't want to see with the debug log level

ShamsUlAzeem: I actually want the users to see when the server is up and running

python/konduit/server.py
ShamsUlAzeem: Now you can override the pid file location as well.

python/konduit/client.py
ShamsUlAzeem: Created a utility for that. The server validation is a little different from client so it wasn't created into a utility.

python/konduit/client.py
ShamsUlAzeem: So as discussed i've separated host and port with host being defaulted to `http://localhost`

python/tests/test_client.py
ShamsUlAzeem: I've implemented the above

Issue #91, opened 2019-11-22
Python runner crashes when input type is NDARRAY and output type is STR


kmotohas:
The face detection example fails after its input type is changed from `STR` to `NDARRAY`.
https://github.com/KonduitAI/konduit-serving/blob/kuzushiji/python/examples/face_detection_pytorch/run_konduit.py
https://github.com/KonduitAI/konduit-serving/blob/kuzushiji/python/examples/face_detection_pytorch/detect_image.py

It seems the server assumes types of input and output are same.

```
21:07:57.699 [vert.x-worker-thread-18] ERROR a.k.s.c.PipelineRouteDefiner - Failed to do inference 
java.lang.ClassCastException: org.datavec.api.writable.Text cannot be cast to org.datavec.api.writable.NDArrayWritable
	at ai.konduit.serving.util.SchemaTypeUtils.toArrays(SchemaTypeUtils.java:263)
	at ai.konduit.serving.executioner.Pipeline.doPipelineArrays(Pipeline.java:108)
	at ai.konduit.serving.executioner.PipelineExecutioner.doInference(PipelineExecutioner.java:288)
	at ai.konduit.serving.configprovider.PipelineRouteDefiner.lambda$null$6(PipelineRouteDefiner.java:333)
	at io.vertx.core.impl.ContextImpl.lambda$executeBlocking$2(ContextImpl.java:316)
	at io.vertx.core.impl.TaskQueue.run(TaskQueue.java:76)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.lang.Thread.run(Thread.java:748)
```

Full error log is found here https://gist.github.com/kmotohas/40de09a8009920d7058f95f5c027ff1b

Inference itself works correctly according to this line https://gist.github.com/kmotohas/40de09a8009920d7058f95f5c027ff1b#file-run_konduit-log-L739

ShamsUlAzeem:
[...]

PR #110, opened 2019-12-01
Makes pipeline step an interface and add validation

agibsonccc:
A bit of refactoring ensuring pipeline step is an interface rather than just an abstract class.


agibsonccc:
@maxpumperla so I guess we should probably think about what we want for validation. Do we want validation mid pipeline as well as at beginning and end? I don't want to be too restrictive.

maxpumperla:
@agibsonccc the PipelineStep interface is good, but it's difficult to understand what else you do. the commit that says "Add Python tests" actually deletes one. where can I see the python test refactoring you mention in the title? sorry, a few old commits blew this up a little.

As to your question, for now I'd like to go for minimizing strange errors. We can loosen restrictions again if we see the need. How would you go about mid-pipe checks? We pass a list of steps to the InferenceConfig, right? Would we then check if "pieces fit together"?

agibsonccc:
@maxpumperla  Yeah I was starting on python tests trying to solve #91 and noticed some bigger picture problems I wanted to solve. Will rename and scope down thanks! 

For mid pipeline validation, I was thinking something simple for now where you just make sure each type can have overlapping input and output types. Then on the end we validate the dataformat.

PR #111, opened 2019-12-01
Pipeline step type validation

agibsonccc:
Pipeline step type validation. This pull requests adds validation upon startup which checks that the  data format is correct or not by checking a list for each pipeline step.

### DIFF
--- a/konduit-serving-api/src/main/java/ai/konduit/serving/pipeline/PipelineStep.java
+++ b/konduit-serving-api/src/main/java/ai/konduit/serving/pipeline/PipelineStep.java
@@ -1,9 +1,12 @@
 package ai.konduit.serving.pipeline;
 
 import  java.io.Serializable;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 
+import ai.konduit.serving.config.Input;
+import ai.konduit.serving.config.Output;
 import org.nd4j.shade.jackson.annotation.JsonSubTypes;
 import org.nd4j.shade.jackson.annotation.JsonTypeInfo;
 import ai.konduit.serving.pipeline.step.*;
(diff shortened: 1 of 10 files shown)


konduit-serving-api/src/main/java/ai/konduit/serving/pipeline/PipelineStep.java
```diff
@@ -27,6 +30,50 @@
...
+
+       return Arrays.stream(validOutputTypes()).anyMatch(input -> dataFormat.equals(input));
+
+   }
+
+
+    /**
+     * Returns the valid {@link Input.DataFormat}
+     * when the input is the beginning of a pipeline
+     * @return
```
maxpumperla: maybe add a note that explains that setting this to `null` is the intended way of allowing any format (for both inputs and outputs)?

konduit-serving-core/src/main/java/ai/konduit/serving/executioner/PipelineExecutioner.java
```diff
@@ -209,6 +212,9 @@ public void init() {
         ServingConfig servingConfig = config.getServingConfig();
         //initialize input and output data types
         this.pipeline = Pipeline.getPipeline(config.getSteps());
+        Preconditions.checkState(config.getSteps().get(0).isValidInputType(servingConfig.getInputDataFormat()),"Configured Serving Configuration input type is invalid");
```
maxpumperla: Can we log the configured type and tell the user which ones are allowed (keeping in mind null means any)? Otherwise this might be difficult to debug, esp. for Python users.

agibsonccc:
@maxpumperla done

maxpumperla:
:+1:

Issue #118, opened 2019-12-03
`konduit init` does not work if whitespace in path 


philip-khor:
```
(base) C:\Users\Skymind AI Berhad>konduit init --os windows-x86_64
Already up to date.
Python 3.7.5
Running command: C:\Users\Skymind AI Berhad\.konduit\konduit-serving\mvnw -Puberjar clean install -Dmaven.test.skip=true -Djavacpp.platform=windows-x86_64  -Dchip=cpu -Ppython -Ppmml
'C:\Users\Skymind' is not recognized as an internal or external command,
operable program or batch file.
Traceback (most recent call last):
  File "C:\Users\Skymind AI Berhad\.konduit\konduit-serving\build_jar.py", line 58, in <module>
    subprocess.run(command, cwd=args.source, shell=True, check=True)
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.7_3.7.1520.0_x64__qbz5n2kfra8p0\lib\subprocess.py", line 512, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command 'C:\Users\Skymind AI Berhad\.konduit\konduit-serving\mvnw -Puberjar clean install -Dmaven.test.skip=true -Djavacpp.platform=windows-x86_64  -Dchip=cpu -Ppython -Ppmml' returned non-zero exit status 1.
```

maxpumperla:
@maxpumperla the git pull bit runs, but mvnw still fails:

Running command: C:\Users\Skymind AI Berhad\.konduit\konduit-serving\mvnw -Puberjar clean install -Dmaven.test.skip=true -Djavacpp.platform=windows-x86_64 -Dchip=cpu -Ppython -Ppmml
Traceback (most recent call last):
  File "C:\Users\Skymind AI Berhad\.konduit\konduit-serving\build_jar.py", line 88, in <module>
    subprocess.call(command)
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.7_3.7.1520.0_x64__qbz5n2kfra8p0\lib\subprocess.py", line 339, in call
    with Popen(*popenargs, **kwargs) as p:
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.7_3.7.1520.0_x64__qbz5n2kfra8p0\lib\subprocess.py", line 800, in __init__
    restore_signals, start_new_session)
[...]

PR #130, opened 2019-12-05
build_jar.py fixes

ShamsUlAzeem:
Checks for the presence of `python/tests` folder. Also, `shell=True` when on windows.

Related to: #118

### DIFF
--- a/build_jar.py
+++ b/build_jar.py
@@ -101,4 +101,4 @@ if __name__ == "__main__":
 
     # Copy the built jar file to the "python/tests" folder if it exists.
     if os.path.isdir(os.path.join(args.source, "python", "tests")):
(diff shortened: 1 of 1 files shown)


build_jar.py
```diff
@@ -95,4 +98,7 @@
         ),
         os.path.join(args.source, args.target),
     )
-    copyfile(args.target, os.path.join(args.source, "python", "tests", args.target))
+
+    # Copy the built jar file to the "python/tests" folder if it exists.
+    if os.path.isdir(os.path.join(args.source, "python", "tests")):
+        copyfile(args.target, os.path.join(args.source, "python", "tests", args.target))
```
philip-khor: do we want to copy from line 99? 

```suggestion
        copyfile(os.path.join(args.source, args.target), os.path.join(args.source, "python", "tests", "konduit.jar"))
```

PR #145, opened 2019-12-08
Reduces amount of method declarations via generics and java 8 default interfaces

agibsonccc:
Reduces the amount of code needed to reproduce type specific builders and allows over riding of 
pipeline step builder logic


konduit-serving-api/src/main/java/ai/konduit/serving/pipeline/PipelineStep.java
```diff
@@ -156,95 +153,106 @@ default boolean isValidPredictionType(PredictionType predictionType) {
      */
     List<String> getInputNames();
 
+
+
     /**
-     * Define a single input for a PipelineStep from column names and types.
-     * The input name will be "default" when using this method.
+     * Define a single input for a TransformProcess Step from explicit
```
maxpumperla: should still be `PipelineStep`, if subclasses inherit this doc string

PR #158, opened 2019-12-13
Python exec autogen tests (WIP)

farizrahman4u:


farizrahman4u:
@agibsonccc @maxpumperla this just generates python scripts, how do we get to tests from here? Right now we only allow 1 config per test class.

maxpumperla:
@farizrahman4u you'd put the tests into the `python/tests` folder in file starting with a `test_` prefix to make `pytest` happy, and build test functions with that prefix as well. also, note that we're using `@pytest.mark` to mark unit and integration tests. use these as suitable, the goal is to keep unit tests in a few-seconds-to-run range.

PR #161, opened 2019-12-16
Updating version to 0.1.0-SNAPSHOT and generating artifacts for konduit (TAR/DEB/RPM)

ShamsUlAzeem:
Updating corresponding modules for generating artifacts for konduit spins for different platforms.


maxpumperla:
@ShamsUlAzeem I still don't think the version number 1.3.0 makes any sense (other than historical / legacy, which nobody should care about).

We could use this time to consider aligning this with DL4J releases (at least roughly). @agibsonccc @wmeddie what do you think?

wmeddie:
@maxpumperla As in aligning it by calling it 1.0.0-beta6 as well?

maxpumperla:
@wmeddie yes, and if patches are needed, do something like `1.0.0-beta6-x`. It'd be either that or starting with something more reasonable like just `0.1`. I don't understand why why would start with `1.3`. That suggests a level of maturity that, frankly, we don't have yet.

ShamsUlAzeem:
@maxpumperla I also want to start with `0.1` or something along the lines as well. Since this hasn't seen any real users yet so we don't know how much we'd have to update our APIs in order to have something stable for general public use.

wmeddie:
I'm not against starting at 0.1.

ShamsUlAzeem:
Alright. Let's see what @agibsonccc thinks about this

agibsonccc:
Go ahead folks.

konduit-serving-rpm/src/main/resources/etc/profile.d/konduit-serving-env.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+# This file defines the environment variables needed to run the konduit serving admin command.
+
+export KONDUIT_SERVING_BASE_ARTIFACT_NAME=konduit-serving
+export KONDUIT_SERVING_DIRECTORY_STRUCTURE=konduit/serving
+export KONDUIT_SERVING_VERSION=0.1.0-SNAPSHOT
+export KONDUIT_SERVING_BACKEND=cpu
```
agibsonccc: We don't have a concept of a backend anymore. Maybe make sure we clean up old assumptions while we're at it here.

konduit-serving-rpm/src/main/resources/etc/profile.d/konduit-serving-env.sh
```diff
@@ -0,0 +1,24 @@
...
+
+export KONDUIT_SERVING_HOME=/opt/${KONDUIT_SERVING_DIRECTORY_STRUCTURE}
+export CONDA_INSTALL_PATH=${KONDUIT_SERVING_HOME}/miniconda
```
agibsonccc: We don't really have the miniconda concept unless this is going to be used for some form of automatic setting of the python path? @maxpumperla can you comment here?

konduit-serving-rpm/src/main/resources/etc/profile.d/konduit-serving-env.sh
ShamsUlAzeem: This isn't an old assumption, these settings are for when we want to use gpu based python libraries or cpu based ones. This is set while we're building with `-Dchip=cpu` or `-Dchip=gpu`

konduit-serving-rpm/src/main/resources/etc/profile.d/konduit-serving-env.sh
ShamsUlAzeem: There's an idea about being able to set the python path automatically using the konduit cli for configurations for python steps. It would look something like this: 
`konduit update-python-path <yaml.config.file>

maxpumperla: @agibsonccc If I understand correctly, Shams wants to have an _optional_ conda installation for users that don't want to rely on their own Python install locally (or don't even have one). in the case of this rpm install it might actually be useful to just ship everything, batteries included. this way you can directly get started with the `konduit` cli without having to do any additional steps (like pip install konduit etc.)

saudet:
What about non-conda users? It sounds like you're tying this to Anaconda. That doesn't sound like a good idea to me.

konduit-serving-rpm/src/main/resources/etc/profile.d/konduit-serving-env.sh
saudet: Including everything doesn't mean using conda. Why does it have to be conda?

ShamsUlAzeem:
@saudet it's supposed to be an optional conda install at the end. Managing it through conda is easier for managing the python projects. 

For non-conda users they can just skip installing the conda package and work with their own (or packaged) jdk or python installs

saudet:
That still doesn't sound right... If we have dependencies on conda packages, then konduit should become a conda package itself. That's the only thing that makes sense.

ShamsUlAzeem:
@saudet that's what I'm doing with conda. We have konduit on conda https://anaconda.org/konduitai/konduit. The python pipeline step is the only step that depends on python packages. Other that that we can just do without python. the only place where the konduit package and its CLI is useful is for serving other steps and we have a pypi module for that.

ShamsUlAzeem:
The conda installation and management through it is for people who just wants everything from scratch (jdk, cuda, python). If they already have some of those installed for them then we'll have optional installs later.

saudet:
Ok, so it sounds to me like you're trying to do an installer for Anaconda + a few additional packages that runs from an RPM. If that's the case, why not just do that instead of doing something that can only install Konduit? Who knows, there might be people who would like to install, I don't know, Jupyter from RPM. One stone two birds kind of thing.

PR #163, opened 2019-12-16
Removing extra dimension from nd4j input adapter

ShamsUlAzeem:
Removing extra dimension from nd4j input adapter


maxpumperla:
@ShamsUlAzeem I'm confused, didn't I do just that in a previous PR or is that another input adapter?

ShamsUlAzeem:
@maxpumperla yup this is for nd4j. the previous one was for numpy input adapter

PR #179, opened 2019-12-22
[WIP]: Multi name json input/output

agibsonccc:
Adds multi name input/output endpoint.


konduit-serving-test/src/test/java/ai/konduit/serving/util/SchemaTypeUtilsTest.java
```diff
@@ -22,16 +22,200 @@
...
+                    break;
+                case Time:
+                    schemaBuilder.addColumnTime(value.name(),TimeZone.getDefault());
+                    break;
+                case Categorical:
+                    schemaBuilder.addColumnCategorical(value.name());
+                    break;
+                case Bytes:
+                    BinaryMetaData binaryMetaData = new BinaryMetaData(value.name());
+                    schemaBuilder.addColumn(binaryMetaData);
```
ShamsUlAzeem: We can probably combine the `Image` and `Bytes` case here

ShamsUlAzeem:
Also, can you write a test for the actual `/dynamicschema` endpoint request?

konduit-serving-test/src/test/java/ai/konduit/serving/util/SchemaTypeUtilsTest.java
agibsonccc: I want to keep bytes and images separate here. Technically an ImageWritable type exists and we should figure out how we want to interact with base64/images.

agibsonccc:
@ShamsUlAzeem we have a doJsonInference (the underlying pipeline execution method that gets invoked) test now that tests all types.  I  want to think about the design of the client side experience first as well as see if there's anything else we need to do for the doJsonInference test coverage.

Issue #182, opened 2019-12-23
Regex in build_jar.py not compatible with 1.0.0-SNAPSHOT 


philip-khor:
Running `konduit build` on latest master returns 

```
Traceback (most recent call last):
  File "C:\Users\Skymind AI Berhad\.konduit\konduit-serving\build_jar.py", line 100, in <module>
    os.path.join(args.source, args.target),
  File "c:\users\skymind ai berhad\appdata\local\continuum\miniconda3\envs\pytorch\lib\shutil.py", line 120, in copyfile
    with open(src, 'rb') as fsrc:
FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Users\\Skymind AI Berhad\\.konduit\\konduit-serving\\konduit-serving-uberjar\\target\\konduit-serving-uberjar-20190722-bin.jar'
```

The regex here: 
https://github.com/KonduitAI/konduit-serving/blob/f0b829ec13e1fc3f675cfc6dbd85dec26f14defe/build_jar.py#L84-L87

searches through `pom.xml` in the root directory for matches, and later on sets the first element of the list `version` as the version number for the source uberjar's file name: 

https://github.com/KonduitAI/konduit-serving/blob/f0b829ec13e1fc3f675cfc6dbd85dec26f14defe/build_jar.py#L93-L101

However, this regex does not match a version number if there are non-digit characters trailing the version number. 

Additionally, the source file's name no longer ends in <version.number>-bin.jar: on my machine, it's 

> konduit-serving-uberjar-0.1.0-SNAPSHOT-custom-windows-x86_64-cpu.jar

philip-khor:
Fixed in #184

PR #184, opened 2019-12-24
Fix build_jar.py (#182)

philip-khor:
- Allow for non-whitespace characters suffix in version number e.g. 1.0.0"-SNAPSHOT"
- remove -bin suffix from JAR filenames in favour of current JAR filename convention: `konduit-serving-uberjar-{version}-{spin.version}-{os}-{arch}.jar`. Assuming `spin.version` is `custom` here.  

Tested on Windows and Ubuntu.

### DIFF
--- a/build_jar.py
+++ b/build_jar.py
@@ -32,18 +32,18 @@ if __name__ == "__main__":
     )
 
     parser.add_argument(
-        "--usePython",
-        type=str,
-        default="true",
-        help="whether to bundle python " "or not (typically not encouraged with arm",
-    )
-
-    parser.add_argument(
-        "--usePmml",
-        type=str,
-        default="true",
-        help="whether to use pmml or not,"
-        " not encouraged if agpl license is an issue",
+        "--spin", 
+        type=str, 
+        default="all", 
+        choices=[
+            "minimal", 
+            "python", 
+            "pmml", 
+            "all"
+        ], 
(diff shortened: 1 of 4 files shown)


build_jar.py
```diff
@@ -95,7 +97,7 @@
             args.source,
             "konduit-serving-uberjar",
             "target",
-            "konduit-serving-uberjar-{}-bin.jar".format(version[0]),
+            "konduit-serving-uberjar-{}-custom-{}-{}.jar".format(version[0], args.os, arch),
```
ShamsUlAzeem: @philip-khor Just for reference, the idea behind a spin version is that we want to have multiple iterations of the built uber jar. For example, if I have `spin.version=minimal` it will build the jar file without the `pmml` and `python` profiles. Similarly, the other types are:

- `spin.version=python` -> add the `python` profile
- `spin.version=pmml` -> add the `pmml` profile 
- `spin.version=all` (keep this as default in `build_jar.py`) -> add `python` and `pmml` profiles

Based on that can you make changes to add an additional `spin` argument that takes care of those profiles internally and additionally add a `-Dspin.version=<spin>` to the build command? 

I see that we have `usePmml` and `usePython` in there as well, you might wanna remove them in order to make this workflow and update documentation as well.

philip-khor: Changes made. I realised CLI (`konduit build` and `konduit init`) need that too. I can go ahead and try making those changes there too.

Issue #181, opened 2019-12-23
error: cannot find symbol for MavenWrapperMain


philip-khor:
I'm getting `error: cannot find symbol` when the MavenWrapperMain class is imported by the Maven Wrapper script for 
- File 
- FileInputStream
- FileOutputStream
- URL
- IOException 
- Authenticator
- ReadableByteChannel
- Channels

See Dockerfile and trace at https://gist.github.com/philip-khor/339df906418a8c0355b1bb0e808f6d9b

This is possibly caused by several import statements removed from [MavenWrapperDownloader.java](https://github.com/KonduitAI/konduit-serving/blob/master/.mvn/wrapper/MavenWrapperDownloader.java) in b247b21. 

```java
import java.net.*;
import java.io.*;
import java.nio.channels.*;
```

This error disappears when `mvn -N io.takari:maven:0.7.7:wrapper` is run before `mvnw`.

philip-khor:
This is not noticeable on certain Linux installations because the `mvnw` script first checks if wget or curl is available; if not it uses the Java script to build the file.

philip-khor:
Closed in #185

PR #185, opened 2019-12-24
Restore imports in MavenWrapperDownloader class

philip-khor:
Also explicitly import required classes. This allows the Maven Wrapper JAR to be downloaded if konduit-serving is built on a Linux system that does not have curl nor wget installed (#181). 

Tested manually by 
- removing maven-wrapper.jar 
- uninstalling curl and wget 
- run `python build_jar.py --os linux-x86_64`


ShamsUlAzeem:
Oh this is awesome!! Thanks @philip-khor

PR #188, opened 2019-12-27
Don't force input_name in .step() method to default

philip-khor:
`input_name` default (`"default"`) already set in function signature. We raise `TypeError` if `input_name` is `None`.

### DIFF
--- a/python/konduit/inference.py
+++ b/python/konduit/inference.py
@@ -148,8 +148,8 @@ def python_step_func(
     output_column_names=None,
     output_types=None,
 ):
-    if input_name is None: 
-        raise TypeError("input_name must not be None")
+    if bool(input_name and input_name.strip()): 
+        raise TypeError("input_name must not be None or empty string")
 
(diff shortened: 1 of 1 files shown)


python/konduit/inference.py
```diff
@@ -148,6 +148,9 @@ def python_step_func(
     output_column_names=None,
     output_types=None,
 ):
+    if input_name is None:
```
ShamsUlAzeem: @philip-khor also make sure it's not `""`

philip-khor:
Testing also for whitespace here, if that's too strict happy to revert to `if not input_name:`

python/konduit/inference.py
```diff
@@ -148,6 +148,9 @@ def python_step_func(
     output_column_names=None,
     output_types=None,
 ):
+    if not bool(input_name and input_name.strip()):
```
philip-khor: Testing also for whitespace here, if that's too strict happy to revert to `if not input_name:`

PR #190, opened 2019-12-31
Redundant "gpu" in filename copied from target

philip-khor:
Copying uberjars from file paths formatted with `args.os` instead of valid JavaCPP platforms results in the following error on GPU builds, because the source filenames are formatted with JavaCPP platform labels, not `os`. Running `konduit build --os windows-x86_64-gpu` results in the following error:

```
Traceback (most recent call last):                                                                    
File "C:\Users\Skymind AI Berhad\.konduit\konduit-serving\build_jar.py", line 105, in <module>
os.path.join(args.source, args.target),
File "c:\users\skymind ai berhad\appdata\local\continuum\miniconda3\lib\shutil.py", line 120, in copyfile                 
with open(src, 'rb') as fsrc:                                                                                       
FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Users\\Skymind AI Berhad\\.konduit\\konduit-serving\\konduit-serving-uberjar\\target\\konduit-serving-uberjar-0.1.0-SNAPSHOT-all-windows-x86_64-gpu-gpu.jar'       
```

This fix tested by running python build_jar.py with os arguments windows-x86_64, windows-x86_64-gpu, linux-armhf


saudet:
That looks like a poor workaround. Why does it do this in the first place? Is it something that needs to be fixed in JavaCPP?

philip-khor:
Shouldn't be a JavaCPP issue. To my understanding (sorry if I misunderstood): 

The user passes one of the following `--os` arguments: 
https://github.com/KonduitAI/konduit-serving/blob/7ac4cda58e4718641ef32b613c1681f2461ec0f7/build_jar.py#L23-L28

which is truncated if it contains "gpu" into a valid JavaCPP platform: 
https://github.com/KonduitAI/konduit-serving/blob/7ac4cda58e4718641ef32b613c1681f2461ec0f7/build_jar.py#L62

The `javacpp.platform` is passed directly to `mvnw` here: 
https://github.com/KonduitAI/konduit-serving/blob/4953e1a1c43ffdc07fc4a9e4afa620a09f7e2f38/build_jar.py#L63-L70

and the jar is built with filename: 

https://github.com/KonduitAI/konduit-serving/blob/f0b829ec13e1fc3f675cfc6dbd85dec26f14defe/konduit-serving-uberjar/pom.xml#L59

I did not notice the `platform` variable earlier, sorry. I'll use that variable instead, but I'm not sure if that makes this fix any better.

saudet:
Hum, why does this script forces us to specify an OS to start with? And why doesn't it make use of the Maven profiles made for that instead of redoing it in Python?
https://github.com/bytedeco/javacpp-presets/wiki/Reducing-the-Number-of-Dependencies
Before starting to rewrite everything in Python, please let me know what is missing from JavaCPP! From what I understand, I don't think anything is missing, so I won't be able to help if you keep all that information to yourself. @ShamsUlAzeem @maxpumperla

wmeddie:
Fixed in my other PR.

PR #194, opened 2020-01-02
Created java based tool for deleting numpy javacpp cache

ShamsUlAzeem:


saudet:
I understand that this is only temporary, but if this is something that becomes important, we should implement this in here: https://github.com/bytedeco/javacpp/blob/master/src/main/java/org/bytedeco/javacpp/tools/CacheMojo.java

ShamsUlAzeem:
Thanks @saudet. Ideally I'll have to totally remove this after the numpy cache issue is solved. The mojo you mentioned looks best for adding it in the future if we wanna keep it for any reason.

PR #196, opened 2020-01-08
[OLD API] Decorator-driven python config files

wmeddie:
This PR should just be a discussion for now about implementing simple decorators for specifying the konduit config.

The goal is to enable users to simply call `konduit serve serve.py` or `python serve.py` on a file like the following:

```python3
import torch
import numpy as np
from model.train import Model
import torchvision.transforms as transforms
from konduit import KonduitServer, serving_function, serving_config
from konduit.types import Image, Json
from typing import Dict

@serving_config(http_port=1337, input_data_format=Image, output_data_format=Json)
class Cifar10Server:
    transform = None
    model = None
    def __init__(self):
        self.transform = transforms.Compose([transforms.ToTensor(),
                                        transforms.Normalize((0.5, 0.5, 0.5), (0.5, 0.5, 0.5))])
        self.model = Model()
        self.model.load_state_dict(torch.load('./model/model.pt'))
        self.model.eval()

    @serving_function(the_prediction=np.ndarray, other_result=float)
    def run(self, the_input: Image, other: float) -> Dict[str, object]:
        input_tensor = self.transform(the_input).unsqueeze(0)
        return { 
            'the_prediction': self.model(input_tensor).detach().numpy(),
            'other_result': other * 42
        }

if __name__ == '__main__':
    KonduitServer.serve(Cifar10Server)
```

We can map the data from the type annotations and decorator arguments to construct the necessary config.json.

This PR is a WIP as it's missing:

- [ ] documentation for all functions
- [ ] Implementation of loading python config files in load.py's server_from_file
- [ ] Unit tests for the new configuration format
- [ ] Not all types are mapped
- [ ] The python runner might need to be extended to support the types still

### DIFF
--- a/python/konduit/decorators.py
+++ b/python/konduit/decorators.py
@@ -1,0 +1,95 @@
+import cv2
+import os
+import functools
+
+from .server import Server
+from .inference import PythonConfig
+from .utils import default_python_path
+
+def serving_config(http_port, input_data_format, output_data_format, log_timings = False):
+    def serving_config_outer(cls):
+        cls._konduit_config_class = True
+        cls._konduit_http_port = http_port
+        cls._konduit_input_data_format = input_data_format
+        cls._konduit_output_data_format = output_data_format
+        cls._konduit_log_timings = log_timings
+
+        @functools.wraps(cls)
+        def serving_config_wrapper(*args, **kwargs):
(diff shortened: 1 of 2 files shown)


ShamsUlAzeem:
Closing due to becoming old api

PR #198, opened 2020-01-09
[OLD API] Ability to create CustomStep from python and yaml

ShamsUlAzeem:
You can create a Custom Step from yaml like this:

`custom_step.yaml` file:
```
serving:
  http_port: 1337
steps:
  custom_step_one:
    type: CUSTOM
    custom_udf_clazz: <class_name>
client:
  port: 1337
```

From the server code side:
```
file_path = "custom_step.yaml"
server = server_from_file(file_path=file_path, start_server=True)
```

Remember to put the jar containing the extra class in the classpath of the server: #197

### DIFF
--- a/konduit-serving-api/src/main/java/ai/konduit/serving/pipeline/CustomStepUDF.java
+++ b/konduit-serving-api/src/main/java/ai/konduit/serving/pipeline/CustomStepUDF.java
@@ -22,14 +22,14 @@
 
 package ai.konduit.serving.pipeline;
 
-import ai.konduit.serving.pipeline.step.CustomPipelineStep;
+import ai.konduit.serving.pipeline.step.CustomStep;
 import org.datavec.api.records.Record;
 
 /**
  * A user defined function for use with the
- * {@link CustomPipelineStep}. This is where
+ * {@link CustomStep}. This is where
  * a user defines an extension for use within a
- * {@link CustomPipelineStep}.
+ * {@link CustomStep}.
  * <p>
  * A user defined function should take in an
  * array of {@link Record} and output an array of {@link Record}
@@ -41,12 +41,12 @@ import org.datavec.api.records.Record;
  *
  * @author Adam Gibson
  */
-public interface CustomPipelineStepUDF {
+public interface CustomStepUDF {
 
 
     /**
      * The user defined function where a user can define custom behavior
-     * for a {@link CustomPipelineStep}
+     * for a {@link CustomStep}
      *
      * @param input the input records (generally 1 record per input name for the pipeline step)
      * @return the output records (generally 1 record per output name for the pipeline step)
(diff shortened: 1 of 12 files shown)


agibsonccc:
@ShamsUlAzeem mind resolving conflicts?

ShamsUlAzeem:
@agibsonccc solved them just now. @wmeddie @kmotohas can you confirm the working of this step through yaml based on your requirements before? IIRC we wanted to make a bert preprocessors with this.

kmotohas:
Is there any existing code that I could refer to for `custom_udf_clazz`?
Is something like JAR file of DataVec's TransformProcess supposed to be `custom_udf_clazz`?

wmeddie:
I added what I *think* it should look like here: https://github.com/wmeddie/custom-konduit-step Can someone give it a shot?

wmeddie:
I think the PR as is is great, but after going through the exercise of making a custom UDF, we need to add some more functionality to UDF configuration before we can expect it to be usable in the real world.  The main issues here:

1. Figuring out what record this UDF is meant to work on.
2. Configuring the UDF.

For 1, the UDF gets a Record[] as input, but it had no idea which record it should be working on (except in the trivial case when there's only 1 record.)  We should pass the same schema configuration to the custom UDF constructor so that the UDF author can use that information to determine the proper index to use for input and output.  (Probably worth adding some helper methods on the base class to make it trivial.)

For 2 I think it'd be best if we can just pass a Map<String, String> to the UDF constructor for every other key inside the json or yaml file that isn't part of the standard schema config.  For example, this is necessary if I have two instances of the same UDF in a pipeline but they should use different vocabulary files.

agibsonccc:
@wmeddie remember, the record array is not actually more than 1 record. More than 1 is actually not possible. It's only 1 record per input name.

wmeddie:
Is this understanding correct?:
the batch -> Record[] (the parameter to the udf method)
the inputs -> records[0].getRecord()
input named "a" -> records[0].getRecord().get(0)
input named "b" -> records[0].getRecord().get(1)

agibsonccc:
No:
"a": record[0]
"b": record[1]

ShamsUlAzeem:
@agibsonccc I believe eduardo is right here... the mapping is done within a single record. In case of the bert network test here: https://github.com/KonduitAI/konduit-serving/blob/c0d550b9935ef2794b77bdba20f8aa007616e68e/python/tests/test_bert_serving.py

the record will look like:
Record[0] = [NDArrayWritable, NDArrayWritable, NDArrayWritable] -> ["IteratorGetNext:0", "IteratorGetNext:1", "IteratorGetNext:4"]

agibsonccc:
@ShamsUlAzeem mind updating this?

AlexDBlack:
@ShamsUlAzeem where is this at?
Seems like a reasonable feature to have, so it would be nice to get this wrapped up...

ShamsUlAzeem:
Hmm have to think about how to add extra configuration options in there. Shouldn't be too hard to do. I'll come to this

ShamsUlAzeem:
Closing due to becoming old api

Issue #209, opened 2020-01-30
Remove assert keyword from Java tests (replace with junit assert methods)


AlexDBlack:
https://github.com/KonduitAI/konduit-serving/blob/master/konduit-serving-test/src/test/java/ai/konduit/serving/pipeline/steps/TransformProcessStepTest.java#L31

Using the Java assert keyword _might_ work for tests, but it's not safe as Java assertions aren't enabled by default. We should use junit assert methods instead

I haven't done a thorough review, there may be more instances of this.

agibsonccc:
Ahh good find! I caught most of those. Thanks for finding this!

AlexDBlack:
https://github.com/KonduitAI/konduit-serving/pull/237

PR #237, opened 2020-02-12
Replacing java asserts with junit assert methods

ShamsUlAzeem:
Fix for #209

### DIFF
--- a/konduit-serving-core/src/main/java/ai/konduit/serving/util/image/NativeImageLoader.java
+++ b/konduit-serving-core/src/main/java/ai/konduit/serving/util/image/NativeImageLoader.java
@@ -37,7 +37,6 @@ import org.datavec.image.data.ImageWritable;
 import org.datavec.image.loader.AndroidNativeImageLoader;
 import org.datavec.image.loader.BaseImageLoader;
 import org.datavec.image.transform.ImageTransform;
-import org.nd4j.base.Preconditions;
 import org.nd4j.linalg.api.concurrency.AffinityManager;
 import org.nd4j.linalg.api.memory.pointers.PagedPointer;
 import org.nd4j.linalg.api.ndarray.INDArray;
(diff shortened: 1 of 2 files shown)


konduit-serving-core/src/main/java/ai/konduit/serving/util/image/NativeImageLoader.java
```diff
@@ -176,13 +177,11 @@ static Mat convert(PIX pix) {
...
                     break;
                 case 8:
-                    pix2 = pix;
-                    break;
                 case 24:
                     pix2 = pix;
                     break;
                 default:
-                    assert false;
+                    Preconditions.checkState(false, String.format("Unrecognized pixel depth of %d", pix.d()));
```
AlexDBlack: Better:
`throw new IllegalStateException("Unrecognized pixel depth of " + pix.d()");`
or
`Preconditions.throwStateEx("Unrecognized pixel depth of %s", pix.d());`

konduit-serving-test/src/test/java/ai/konduit/serving/pipeline/steps/TransformProcessStepTest.java
```diff
@@ -28,17 +27,16 @@ public void testTransformProcessPipe() throws Exception {
...
-        assert tpps.getOutputNames().contains("bar");
+        assertTrue(tpps.getOutputNames().contains("foo"));
+        assertTrue(tpps.getOutputNames().contains("bar"));
 
         TransformProcessStep tpps2 = new TransformProcessStep(tp, outputSchema);
 
-        assert tpps2.getTransformProcesses().get("default") == tp;
-
-        assert Arrays.equals(tpps2.getInputSchemas().get("default"), SchemaTypeUtils.typesForSchema(inputSchema));
+        assertSame(tpps2.getTransformProcesses().get("default"), tp);
```
AlexDBlack: do we want assertSame, or assertEquals here?
assertSame -> must be exact same object (and hence by definition also equal)
assertEquals -> can be different object, but still has to be equal

ShamsUlAzeem: I think assertSame would make more sense since what we're trying to check is that if the constructor and the builder method are assigning the same object to the different input names for steps.

Issue #240, opened 2020-02-12
Use SLF4J for logging


AlexDBlack:
Currently we have exclusions for SLF4J all throughout Konduit Serving, for example in core:
https://github.com/KonduitAI/konduit-serving/blob/master/konduit-serving-core/pom.xml

The tests at least use logback classic directly, not through slf4j.

Issue #220, opened 2020-01-30
Add JSON/YAML test coverage tracking for config classes


AlexDBlack:
The idea: we introduce an interface: "Jsonable" or something (but with a better name).
This interface would have toJson() and toYaml() methods, which are useful/good irrespective of test coverage.

For testing, we scan the classpath for everything that implements this interface, and then ensure that we have at least one JSON test and one YAML test for each; tests should fail automatically if we introduce a new config class that doesn't have a corresponding JSON/YAML ser/de test.

As we add more classes, it will be hard to ensure that every class is tested for to/from JSON and YAML unless we have some sort of tooling like this..

Issue #235, opened 2020-01-31
Assorted smaller issues to be fixed


AlexDBlack:
Many of these can probably be fixed together, combining here to avoid more issue spam for small issues

1. PipelineStepRunner destroy() method - might as well make this close() and have the class implement closeable interface
2. ModelConfig should (probably) be an abstract class, with concrete subtypes
3. TensorDataTypesConfig uses `org.nd4j.tensorflow.conversion.TensorDataType` not `ai.konduit.serving.model.TensorDataType`
4. PipelineStepRunner should just have a way to get the PipelineStep, instead of having inputTypes() and outputTypes() methods
5. PipelineRouteDefiner - I don't believe we have a summary of the endpoints listed here in the docs, and I don't recall seeing some of these at all in the docs
6. Profile check error message is bad during build - I only get "You must set a chip property!" if I don't set a chip via -D (initially I tried profiles, -P).
7. We should consider adding .npz support, for anywhere users need to feed in multiple NumPy arrays
8. Additional validation - for example TF placeholder datatypes vs. KS configuration

Issue #229, opened 2020-01-31
Server should report useful information when starting


AlexDBlack:
Looking at the examples, the server doesn't report much of anything when it starts.

https://github.com/KonduitAI/konduit-serving-examples/blob/master/notebooks/02_tensorflow-basic-bert.ipynb

```
server.start()
Starting server.................

Server has started successfully.
```

We should at least report the port used, and some basic information about the number and type of pipeline steps. Maybe also a briefly list the endpoints available, or what the user should read to set up a client to connect to this server.
The logs directory for this server is also something important that we should report.

Not a big deal, but it could result in better UX.

PR #244, opened 2020-02-18
Various fixes, cleanup and refactoring

AlexDBlack:
Fixes: https://github.com/KonduitAI/konduit-serving/issues/220
Fixes: https://github.com/KonduitAI/konduit-serving/issues/240

Fixes 1-4 of: https://github.com/KonduitAI/konduit-serving/issues/235
Partial fix for: https://github.com/KonduitAI/konduit-serving/issues/229

Note that from now on, anything that should be convertable to JSON/YAML in Java should implement the TextConfig interface. This is for a couple of reasons:
(a) It allows coverage tracking - i.e., if we add a new config class, and it doesn't have tests (for JSON/YAML conversion), we'll automatically know
(b) Adds default toYaml and toJson methods

Also introduces DL4JConfig and KerasConfig instead of having these use the (now abstract) ModelConfig class.
I haven't (yet) modified anything on the Python side, so there may be some more changes required.


ShamsUlAzeem:
@AlexDBlack so hopefully you won't have to modify anything on the python side (for configurations) as it's generated by the CodeGen for python. https://github.com/KonduitAI/konduit-serving/blob/master/konduit-serving-codegen/src/main/java/ai/konduit/serving/codegen/pythoncodegen/CodeGen.java

konduit-serving-core/src/main/java/ai/konduit/serving/pipeline/steps/InferenceExecutionerStepRunner.java
```diff
@@ -108,27 +108,13 @@ public InferenceExecutionerStepRunner(PipelineStep pipelineStep) {
...
-    public Map<String, SchemaType[]> outputTypes() {
-        Map<String, SchemaType[]> ret = new LinkedHashMap<>();
-        for (int i = 0; i < pipelineStep.getOutputNames().size(); i++) {
-            ret.put(pipelineStep.getOutputNames().get(i), new SchemaType[]{SchemaType.NDArray});
-        }
-
-        return ret;
+    public PipelineStep<?> getPipelineStep() {
+        return pipelineStep;
     }
```
ShamsUlAzeem: Do we need this considering we have it implemented in the BaseStepRunner?

konduit-serving-test/src/test/java/ai/konduit/serving/config/ConfigJsonCoverageTrackingTests.java
```diff
@@ -0,0 +1,271 @@
...
+
+    @Test
+    public void testJsonExpanderTransformStep(){
+        testConfigSerDe(JsonExpanderTransformStep.builder()
+                .inputNames(Arrays.asList("x", "y"))
+                .outputNames(Arrays.asList("z","a"))
+                .inputSchema("x", new SchemaType[]{SchemaType.NDArray, SchemaType.Boolean})
+                .outputSchema("z", new SchemaType[]{SchemaType.Image, SchemaType.Categorical})
+        .build());
+    }
```
ShamsUlAzeem: There's some weird behavior with this test. The [toJson](https://github.com/KonduitAI/konduit-serving/pull/244/files#diff-fe2de6c522278bba41f95729aa38d514R94) method is adding a whitespace character at the end of `ai.konduit.serving.pipeline.step.JsonExpanderTransformStep<JsonExpanderTransformStep()>`

ShamsUlAzeem: `c.equals(fromJson);` is giving false for some reason. Is this because of `@EqualsAndHashCode(callSuper = true)`

ShamsUlAzeem:
Maybe a custom annotation is better to add those fromJson and fromYaml methods. But I'm not sure if it's going to be picked up by the IDE. 

Also, maybe the test can be parametrized cuz it's easier to see individual results. Haven't gone through it in detail but something to consider

konduit-serving-core/src/main/java/ai/konduit/serving/pipeline/steps/InferenceExecutionerStepRunner.java
AlexDBlack: No, we don't. I'll remove it.

konduit-serving-test/src/test/java/ai/konduit/serving/config/ConfigJsonCoverageTrackingTests.java
AlexDBlack: Looks like this is a bug/limitation of Java/Lombok, for arrays in generic collections - lombok just uses Object.equals, and so does HashMap.

I can reproduce it simply here:
https://gist.github.com/AlexDBlack/3a36cca09f30367023fe2e1a1e230760

So we need a custom equals method, easy enough fix.

AlexDBlack: I ended up refactoring the use of arrays to lists, so we don't end up running into this sort if equality issue again anywhere else.

AlexDBlack:
> Maybe a custom annotation is better to add those fromJson and fromYaml methods. But I'm not sure if it's going to be picked up by the IDE.

That's a good idea and I looked into that, but it seems like we can't generate static methods with annotations.
Well, we can, but it looks like we'd need a Maven (and hence IDE) plugin to do so (that's how things like Lombok `@Builder` annotation generating a static `builder()` method get around this.


As for parameterized tests, no, I think I'll keep it the way it is.
IMO it wouldn't add anything here, other than making the tests a little more complex.

Issue #228, opened 2020-01-31
CLI and Python API: Add automatic/random port selection


AlexDBlack:
A minor thing, not a high priority.

Currently if users want a random port, they need to do something like this (for the Python API):
```
port = np.random.randint(1000, 65535)
```
Which was taken directly from here: https://github.com/KonduitAI/konduit-serving-examples/blob/master/notebooks/01_tensorflow-basic-mnist.ipynb

Of course there's no guarantee that the randomly selected port is actually free/available.

Users should just be able to set port 0 or -1 or (1000-65535) or something and have KS select a random free port, reporting the selected port when they start the server.

PR #254, opened 2020-02-26
Port conflict resolution and tests

### LINKED ISSUE #229: Server should report useful information when starting

### LINKED ISSUE #246: Tests: free port selection can (sometimes) be unreliable?
2 runs of this PR, with no code changes: https://github.com/KonduitAI/konduit-serving/pull/244
First run failed with this:
```
[2020-02-20T04:22:06.242Z] [INFO] Running ai.konduit.serving.verticles.samediff.SameDiffVerticleNumpyTest
[2020-02-20T04:22:06.242Z] 04:22:06.136 [vert.x-worker-thread-0] INFO  a.k.s.c.PipelineRouteDefiner - Using metrics registry io.micrometer.prometheus.PrometheusMeterRegistry for inference
[2020-02-20T04:22:07.174Z] 04:22:07.152 [vert.x-worker-thread-0] INFO  a.k.s.c.PipelineRouteDefiner - Using metrics registry io.micrometer.prometheus.PrometheusMeterRegistry for inference
[2020-02-20T04:22:08.546Z] 04:22:08.197 [vert.x-worker-thread-2] INFO  a.k.s.v.inference.InferenceVerticle - Server started on port 33071 with 1 pipeline steps
[2020-02-20T04:22:08.546Z] [ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 2.11 s <<< FAILURE! - in ai.konduit.serving.verticles.samediff.SameDiffVerticleNumpyTest
[2020-02-20T04:22:08.546Z] [ERROR] ai.konduit.serving.verticles.samediff.SameDiffVerticleNumpyTest.runAdd  Time elapsed: 2.108 s  <<< ERROR!
[2020-02-20T04:22:08.546Z] java.net.ConnectException: Connection refused (Connection refused)
[2020-02-20T04:22:08.546Z] 	at ai.konduit.serving.verticles.samediff.SameDiffVerticleNumpyTest.runAdd(SameDiffVerticleNumpyTest.java:127)
```

Second run (simple restart for CI run) passed.

Test passes locally also.

ShamsUlAzeem:
Fixes #229, #246 and #228 

This PR also looks at forwarding the exceptions to the main deploy method at https://github.com/KonduitAI/konduit-serving/blob/bde2e9f94f8fc8806b42d56ac2339898710864a4/konduit-serving-core/src/main/java/ai/konduit/serving/configprovider/KonduitServingMain.java#L131

Had to add the `@SneakyThrows` annotation to bypass  some of the compilation checks 

Looks something like this in action:
```
19:53:30.214 [vert.x-worker-thread-0] INFO  a.k.s.m.l.t.TensorflowGraphHolder - Loading graph
Feb 26, 2020 7:53:39 PM ai.konduit.serving.configprovider.KonduitServingMain
INFO: Deployed verticle ai.konduit.serving.verticles.inference.InferenceVerticle
19:53:39.817 [vert.x-eventloop-thread-3] INFO  a.k.s.v.inference.InferenceVerticle - Server started on port 53119 with 1 pipeline steps  <-----------this is from java which is read by python subprocess


Server has started successfully on port 53119 <-------- the port is mentioned here

[[9.9814737e-01 1.8526246e-03]
 [7.1787997e-04 9.9928218e-01]
 [9.9910551e-01 8.9446508e-04]
 [9.9260944e-01 7.3905047e-03]]

Process finished with exit code 0
```


konduit-serving-api/src/main/java/ai/konduit/serving/util/PortUtils.java
```diff
@@ -0,0 +1,23 @@
+package ai.konduit.serving.util;
+
+import lombok.extern.slf4j.Slf4j;
+
+import java.io.IOException;
+import java.net.ServerSocket;
+
+@Slf4j
```
agibsonccc: We don't use slf4j here I believe? Were we supposed to be?

ShamsUlAzeem: Ah yeah sorry about that. There were a couple of other functions which were used to check if the ports were in use or not. Didn't end up needing them. They were using the logs

ShamsUlAzeem: https://github.com/KonduitAI/konduit-serving/pull/254/commits/f4f54cfaf15f3b81c17a2e7ce4a7897998986067

konduit-serving-api/src/main/java/ai/konduit/serving/util/PortUtils.java
```diff
@@ -0,0 +1,20 @@
+package ai.konduit.serving.util;
+
+import java.io.IOException;
+import java.net.ServerSocket;
+
+public class PortUtils {
```
AlexDBlack: Copyright header, also add `@NoArgsConstructor`

konduit-serving-api/src/main/java/ai/konduit/serving/util/PortUtils.java
```diff
@@ -0,0 +1,20 @@
...
+     * @return single available port number
+     */
+    public static int getAvailablePort() {
+        try {
+            try (ServerSocket socket = new ServerSocket(0)) {
+                return socket.getLocalPort();
+            }
+        } catch (IOException e) {
+            throw new IllegalStateException("Cannot find available port: " + e.getMessage(), e);
+        }
```
AlexDBlack: We can do this with a single try/catch
```
        try (ServerSocket socket = new ServerSocket(0)) {
             return socket.getLocalPort();
        } catch (IOException e) {
            throw new IllegalStateException("Cannot find available port: " + e.getMessage(), e);
        }
```

konduit-serving-api/src/main/java/ai/konduit/serving/verticles/base/BaseRoutableVerticle.java
```diff
@@ -51,7 +58,8 @@ public BaseRoutableVerticle() {
      * Start an http server the port with the value configured
      * as the httpPort key found in {@link #config()}
      */
-    protected void setupWebServer() {
+    @SneakyThrows
+    protected void setupWebServer() throws Throwable {
```
AlexDBlack: Do we need both `@SneakyThrows` _and_ `throws Throwable`? The former should make the latter unnecessary?

konduit-serving-core/src/main/java/ai/konduit/serving/verticles/inference/InferenceVerticle.java
```diff
@@ -99,28 +105,57 @@ public void init(Vertx vertx, Context context) {
...
-            String portEnvValue = System.getenv(VerticleConstants.PORT_FROM_ENV);
-            if (portEnvValue != null) {
-                portValue = Integer.parseInt(portEnvValue);
+    protected void setupWebServer() throws Throwable {
+        String portEnvValue = System.getenv(VerticleConstants.PORT_FROM_ENV);
+        if (portEnvValue != null) {
+            try {
+                port = Integer.parseInt(portEnvValue);
+            } catch (NumberFormatException e) {
+                log.error("Environment variable value isn't a valid port number",  e);
```
AlexDBlack: Let's actually log what they provided for easier debugging?
```
log.error("Environment variable value isn't a valid port number: \"{}\"", portEnvValue, e);
```

konduit-serving-core/src/main/java/ai/konduit/serving/verticles/inference/InferenceVerticle.java
```diff
@@ -99,28 +105,57 @@ public void init(Vertx vertx, Context context) {
...
-                    if (listenResult.failed()) {
-                        log.error("Could not start HTTP server on port {}", portValueFinal);
-                        listenResult.cause().printStackTrace();
+                .listen(port, inferenceConfiguration.getServingConfig().getListenHost(), result -> {
+                    if (result.failed()) {
+                        Throwable cause = result.cause();
+
+                        if(cause instanceof BindException && cause.getMessage().equalsIgnoreCase("Address already in use")) {
+                            log.info("Port: {} is already in use. Trying another port...", port);
+                            startVertxServer(0, nSteps);
```
AlexDBlack: Couple of things here
(a) we should probably only retry with a new port if the user wants a random port. If they want a _specific, fixed port_ (as set in their config) we should just propagate an exception instead of giving them something different
(b) Let's add some sort of "maximum number of attempts" to this? Suppose something is really wrong with the system, we don't want KS trying indefinitely if it can never succeed

konduit-serving-test/src/test/java/ai/konduit/serving/util/PortConflictsResolutionTest.java
```diff
@@ -0,0 +1,163 @@
+package ai.konduit.serving.util;
```
AlexDBlack: Copyright header

python/tests/test_json.py
```diff
@@ -11,6 +11,9 @@
 
 
 @pytest.mark.unit
+@pytest.mark.skip("Can't seem to find the toJson method inside the inference configuration. This might be due to the "
+                  "TextConfig interface addition")
+# TODO: fix this test
```
AlexDBlack: What's the problem here? Are we trying to indirectly use Java reflection or something from Python to find a specific method?

Also, if there was a problem introduced by my PR that added the TextConfig interface, why wouldn't it show up in either my local testing or CI tests? Are python tests not being run as part of the maven build/test?

konduit-serving-api/src/main/java/ai/konduit/serving/util/PortUtils.java
ShamsUlAzeem: Sorry about that. Keep forgetting about that...

konduit-serving-api/src/main/java/ai/konduit/serving/verticles/base/BaseRoutableVerticle.java
ShamsUlAzeem: We don't, removing the `throws Throwable`

ShamsUlAzeem: Hmm actually I think I'll keep the latter and remove the former one

python/tests/test_json.py
ShamsUlAzeem: Not sure what's wrong here exactly but the toJson method wasn't found. Also, it could see that the classes weren't initialized correctly with the fromJson method. Maybe it's something specific with pyjnius. 

Also, yes we don't run the python tests as part of the maven tests. CI doesn't have this setup as well. To run it manually, we'll have to install a local anaconda distro and run the `build_jar.py` script in the root as `python build_jar.py --os=<platform>` and then go to `python/tests` and run `pytest .`

ShamsUlAzeem: If we make this part of the maven build/test we'll still have to install a python distro (anaconda preferred). Pinging @sshepel

ShamsUlAzeem: We can decide something here and then make it part of the CI

konduit-serving-core/src/main/java/ai/konduit/serving/verticles/inference/InferenceVerticle.java
ShamsUlAzeem: So, the maximum number of attempts didn't make sense to me after we don't want to do automatic port resolution. Now it will just throw an error if it isn't able to start the server on the first go. The port number `0` is responsible for automatic port resolution

python/tests/test_json.py
AlexDBlack: Thanks @ShamsUlAzeem - cc @sshepel 
We definitely need this as part of the maven build, and running as part of the CI, ASAP - we can't have a situation where PRs are (unknowingly, like in the case of my PR) able to break the build even though "mvn test" and CI says they are fine.

https://github.com/KonduitAI/konduit-serving/issues/257
https://github.com/KonduitAI/konduit-serving/issues/258

konduit-serving-test/pom.xml
```diff
@@ -149,6 +147,11 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>com.github.stefanbirkner</groupId>
+            <artifactId>system-rules</artifactId>
+            <version>1.19.0</version>
+        </dependency>
```
AlexDBlack: Is this used anywhere? (I can't see any imports/changes referencing this)
If it is used: it should be test scope, I think, based on the description?
If it's not used: remove it

python/tests/test_json.py
AlexDBlack: Hm... as for why toJson isn't found - my guess would pyjnius not liking the java 8 default method on the interface?
https://github.com/KonduitAI/konduit-serving/blob/master/konduit-serving-api/src/main/java/ai/konduit/serving/config/TextConfig.java#L30-L46

I've opened an issue so we don't forget: https://github.com/KonduitAI/konduit-serving/issues/259

konduit-serving-test/pom.xml
ShamsUlAzeem: Ah yes this is used with this import: https://github.com/KonduitAI/konduit-serving/blob/6c6e709e67fc6c3f64395d2b4eb77d63c8199cf2/konduit-serving-test/src/test/java/ai/konduit/serving/util/PortsTest.java#L44.

I've added the test scope

PR #256, opened 2020-02-27
Logging endpoints and tests

### LINKED ISSUE #217: Add the option to expose logging (and stdout/stderr) at a REST endpoint
This should be an option, perhaps disabled by default.

Suppose I've deployed a model on a remote machine, and I've run into some sort of issue.
Currently, I'll need to ssh into that machine, find where the logs for this particular model server are, and read through them to debug the issue.
It would be nice if I could see the logs (and stdout/stderr) directly at an endpoint, like `/logs`.

### LINKED ISSUE #226: Capture stdout / stderr to log files
At present, we rely on Java logging (slf4j). This works quite well for Java code, but it doesn't capture logging/printing from:
* Python code
* C/C++ code (native libraries)
In KS, we'll likely have plenty of both.

If possible, we should capture the whole process stdout/stderr to a file, instead of just capturing java logging only.

ShamsUlAzeem:
Fixes #217. Here's how it looks like right now: 

Default logs page.
Screenshot 2020-03-02 at 11 39 50 PM

Last 10 logs (/logs/10)
Screenshot 2020-02-27 at 11 31 50 PM

All Logs (/logs/all)
Screenshot 2020-02-27 at 11 32 14 PM

Stuff remaining to do:
- [x] Add support for stdout/stderr
- [x] Download button for all logs
- [x] Endpoint for downloading logs as a zip file
- [x] ~Separate out log file naming based on deployment ID~ Keeping everything in `main.log` for now. You can configure it with `KONDUIT_SERVING_LOG_DIR` environment variable in which case it will be inside `${KONDUIT_SERVING_LOG_DIR}/main.log`, otherwise it will be inside `${user.dir}/main.log` where `user.dir` is the system property from where the java process was run from...


ShamsUlAzeem:
@AlexDBlack a few hiccups while download the files as text or zip but the rest is ready for review

konduit-serving-core/pom.xml
```diff
@@ -278,6 +278,12 @@
             <artifactId>logback-classic</artifactId>
             <version>${logback.version}</version>
         </dependency>
+
+        <dependency>
+            <groupId>uk.org.lidalia</groupId>
+            <artifactId>sysout-over-slf4j</artifactId>
+            <version>1.0.2</version>
+        </dependency>
```
AlexDBlack: I wasn't aware of this, it's cool.
MIT/X11 license should be fine for us too.
Let's put version into parent pom properties.

konduit-serving-core/src/main/java/ai/konduit/serving/configprovider/PipelineRouteDefiner.java
```diff
@@ -212,6 +224,74 @@ public Router defineRoutes(Vertx vertx, InferenceConfiguration inferenceConfigur
...
+                    .handler(ctx -> {
+                String numberOfLinesString = ctx.pathParam("numberOfLastLinesToRead");
+
+                try {
+                    ctx.response().putHeader(HttpHeaders.CONTENT_TYPE, "text/plain").end(LogUtils.getLogs(Integer.parseInt(numberOfLinesString)));
+                } catch (Exception e) {
+                    if(e instanceof NumberFormatException) {
+                        if("download".equalsIgnoreCase(numberOfLinesString)) {
+                            ctx.response().sendFile(LogUtils.getLogsFile().getAbsolutePath()).end();
+                        } else if("downloadAsZip".equalsIgnoreCase(numberOfLinesString)) {
```
AlexDBlack: Let's clean this up a bit... don't try to parse integer and catch exceptions
Instead, something like:
```
response = ctx.response().putHeader(HttpHeaders.CONTENT_TYPE, "text/plain");
if(numberOfLinesString.matches("\\d+"){ //Or maybe "-?\\d+" if you want negatives allowed here
    response.end(...)
} else if("download".equalsIgnoreCase(numberOfLinesString)) {

} else ...
```
```

File: konduit-serving-core/src/main/java/ai/konduit/serving/configprovider/PipelineRouteDefiner.java [hunk 3]
```diff
@@ -212,6 +224,74 @@ public Router defineRoutes(Vertx vertx, InferenceConfiguration inferenceConfigur
...
+                                while((length = fis.read(bytes)) >= 0) {
+                                    zipOut.write(bytes, 0, length);
+                                }
+
+                                fis.close();
+                                zipOut.close();
+                                fos.close();
+
+                                ctx.response().sendFile(zipFile.getAbsolutePath()).end();
+                            } catch (Exception exception) {
```
AlexDBlack: Use try-with-resources here. If anything goes wrong before the `X.close()` calls we'll leave a bunch of open files etc.

Also use buffered input and output streams whenever dealing with files.

Also manual buffer handling in a loop should be unnecessary... things like commons io - IOUtils.copy(InputStream, outputStream)

konduit-serving-core/src/main/java/ai/konduit/serving/util/LogUtils.java
```diff
@@ -0,0 +1,146 @@
...
+    /**
+     * Reads the last n lines from a file
+     * @param file file where the data to be read is.
+     * @param numOfLastLinesToRead the number of last lines to read
+     * @return read lines
+     */
+    public static String readLastLines(File file, int numOfLastLinesToRead) {
+        List<String> result = new ArrayList<>();
+
+        try (ReversedLinesFileReader reader = new ReversedLinesFileReader(file, Charset.defaultCharset())) {
```
AlexDBlack: Let's make sure logback is configured for UTF8 and use UTF8 here also - StandardCharsets.UTF8
I'm worried about windows - which defaults to windows 1252 encoding (according to `[System.Text.Encoding]::Default` in powershell), which is _possibly_ what Charset.defaultCharset() will return too.
Better to force utf8 everywhere.

konduit-serving-core/src/main/java/ai/konduit/serving/util/LogUtils.java
```diff
@@ -0,0 +1,146 @@
...
+            String line;
+            while ((line = reader.readLine()) != null && result.size() < numOfLastLinesToRead) {
+                result.add(line);
+            }
+        } catch (IOException e) {
+            e.printStackTrace();
+            return "";
```
AlexDBlack: Should we return an error message and the stack trace instead of empty string?
Suppose a user is trying to get logs via the UI, and something goes wrong. At present, they'll just get empty string in UI, which could be quite confusing for them.

konduit-serving-core/src/main/java/ai/konduit/serving/util/LogUtils.java
```diff
@@ -0,0 +1,146 @@
...
+        if(logsFile == null || !logsFile.exists()) return "";
+
+        if(numOfLastLinesToRead > 0) {
+            return readLastLines(logsFile, numOfLastLinesToRead);
+        } else {
+            try {
+                return FileUtils.readFileToString(logsFile, Charset.defaultCharset());
+            } catch (IOException e) {
+                log.error("Error reading file: ", e);
+                return "";
```
AlexDBlack: As above re: defaultCharset and what we return for error

ShamsUlAzeem:
@farizrahman4u can you look at the test case failure here: https://jenkins.konduit.ai/blue/organizations/jenkins/konduitai%2Fkonduit-serving/detail/PR-256/7/pipeline/136#step-142-log-1101

ShamsUlAzeem:
@AlexDBlack seems like there's a test failure introduced through snapshots. It's inconsistent while reproducing. Other than that, I've taken care of the review comments.

AlexDBlack:
I've triggered tests again, now that we should have new snapshots up...
Looks like there's still an issue:
```
[2020-03-20T10:12:33.199Z] [ERROR]   TestPythonJsonNdArrayScalarInput.testInferenceResult:117 2 expectations failed.

[2020-03-20T10:12:33.199Z] Expected status code <200> doesn't match actual status code <500>.
```

ShamsUlAzeem:
Yeah these were are fixed over here. https://github.com/KonduitAI/deeplearning4j/pull/330. I've asked @farizrahman4u to look at the failing tests over there.

Issue #210, opened 2020-01-30
CLI: don't require config for predict, stop etc methods


AlexDBlack:
Providing the config as an _option_ is fine.
But requiring it every time the user wants to do something isn't great UX IMO.

I'm referring to the --config arg in CLI:
`konduit predict-numpy --config ../yaml/konduit.yaml ...`
`konduit stop-server --config ../yaml/konduit.yaml ...`
https://github.com/KonduitAI/konduit-serving-examples/blob/master/notebooks/00a_serve-from-yaml_cli.ipynb

What would be much nicer for users: a docker-style "server name" way of identifying it.
After launching a container, you can refer to it by name (i.e., this thing - https://frightanic.com/computers/docker-default-container-names/ ) or by ID.

Currently what we have is like asking the user to provide the dockerfile whenever they want to do anything.

agibsonccc:
@AlexDBlack  generally this would be an orchestration extension. Right now, this would mean storage of configuration state on the backend which we don't currently have.

PR #263, opened 2020-03-10
Eliminating the use of --config with every konduit-serving python cli command

ShamsUlAzeem:
I would prefer we create a cli using vertx Launcher and extending it. But this is the simpler solution for #210


AlexDBlack:
@ShamsUlAzeem Given our Vertx CLI plans - do we merge this (once tests are passing)?

ShamsUlAzeem:
Yup I think so. This one should do until we have the new CLI done

PR #268, opened 2020-03-18
Add ADRs for CLI and Client REST APIs

ShamsUlAzeem:
This adds architecture design records for discussing changes related to konduit-serving CLI and rest client APIs


treo:
Some notes on the wording of the ADRs:
Please use a more direct language. The first paragraph of ADR 0001 would then be:

> We create KonduitServingNodeConfigurer as an extension of the Vert.x Launcher class. It will 
> contain the  extra code that's available in both of the above classes (KonduitServingMain and 
> KonduitServingNodeConfigurer).  We extend the CLI by creating seperate classes for every 
> command we want to add (see the documentation here). This way it is possible to register and 
> unregister commands in the Vert.x Launcher individually. 


The biggest difference between the original way you wrote it and my suggested version is that my version states what we are doing. In contrast, your version says what we can do. The difference here is important, because after the proposal changes its state to accepted, it has become a decision. The decision is supposed to be about what we are doing, not about what we can do. 


Concerning the contents of the ADRs, I'll let @AlexDBlack comment on that.

ShamsUlAzeem:
@treo Thanks for the suggestions. I'll make changes to it when I have final comments from Alex as well

AlexDBlack:
Thanks @Paul for the comments, I agree there ("can do x" is potentially ambiguous (is it part of proposal or not?); "will do x" is not ambiguous).

Feedback for first of 2 ADRs: 0001-Extend_vertx_launcher_for_CLI

On the whole I like this proposal, sounds like a good proposal to me.

Couple of questions:

1. How would running KS from Python work?
ATM part of the Python CLI is for building the JAR...
With this approach, I guess we restrict the Python CLI side of things to doing 2 things:
(a) detecting if JAR is present, and building it if necessary (maybe in the future downloading it instead)
(b) passing all commands - including "--help" etc - to the JAR file

That's certainly nicer and more maintainable than having 2 separate CLIs (or only having a proper one in Python, or whatever).
Is that how you imagined this working?


2. "--instances 3" - how does this work in practice?
Does it mean 3 separate java processes, all with the same config, and client talks to load balancer at the port they specified, (load balancer hands of work to the 3 instances)?

AlexDBlack:
Feedback for second ADR: 0002-Use_swagger_annotations_for_generating_client_APIs.md


OpenAPI being able to generate clients for a ton of languages is definitely a huge advantage
looking at it, we have all languages we might want to support - C?, C#, C++?, Go?, Java, Kotlin, JS, Python, R?, Scala, Swift, Typescript
I can't imagine us writing and maintaining all of those manually. Though there will be at least _some_ manual polishing on to of the generated code, as you point out...


> Documentation and maintenance will be easier.

I know we can generate docs for the endpoints - but how will this integrate with GitBook?


> We'll strictly have to adhere to how openapi wants us to define our APIs so, we'll lose a bit of flexibility there.

Can you elaborate on this - what limitations/restrictions might we face?


> Since some of the generated APIs might be lacking so we'll have to write a small layer of wrapper APIs on top of the generate code and docs. So, that will take additional time.

Right, like if we want numpy/INDArray types or whatever in our client APIs, I'm assuming that's manual - there's no way to tell the openapi generator that you want to use numpy arrays or whatever...

ShamsUlAzeem:
@AlexDBlack 

#### 0001-Extend_vertx_launcher_for_CLI.md

> How would running KS from Python work?

It's exactly how you've described it. We'll make sure that we can detect the JAR file based on environment variables and if it's not available we'll download it. We'll have to make sure that the version that's corresponding to the python package is the one that we download. 

Secondly, yeah we'll pass all the commands to the java version of the CLI. Even the --help command will be written in java. 

> "--instances 3" - how does this work in practice?

This is pretty straightforward. Vertx handles this internally. That `--instances` option is internal to vert.x. It can run multiple verticles on the same port and runs then in the same java process. It separate them into multiple threads. I've a small repo [here](https://github.com/ShamsUlAzeem/VertxMetricsDemonstrator/blob/master/src/main/java/tests/shamsulazeem/VerticleMetricsDemonstrator.java) that demonstrate how it works and looks like in action. This is not exactly running on multiple nodes. It's just a way to utilize multiple cpu threads.

#### 0002-Use_swagger_annotations_for_generating_client_APIs.md

> I know we can generate docs for the endpoints - but how will this integrate with GitBook?

Gitbook doesn't have a way to integrate open api specification like readme [did](https://preview.readme.io/) but it's [open-source](https://github.com/readmeio/api-explorer). We can use that to showcase our REST APIs. Other way we can do is to generate [static html](https://raw.githack.com/swagger-api/swagger-codegen/master/samples/html/index.html) pages and update them through CSS. Another option is to use [swagger-ui](https://petstore.swagger.io/?_ga=2.10633204.1573157324.1584626439-1599309769.1584031022)

> Can you elaborate on this - what limitations/restrictions might we face?

One thing is not being able to freely define functions and their overloads since we'll have to make sure they fit together nicely with swagger annotations and also if the generated apis in different languages can easily adapt our API framework.Also, we can't use all kind of objects since some of those won't make sense for openapi. For example, using `Map<String, File>`. This won't work for us to send multipart requests since swagger expects a `File[]` instead to take care of that (these are some small things we'll have to be wary of). So, one of the things we'll have to do is that the already generated API is easily extendable. 

> Right, like if we want numpy/INDArray types or whatever in our client APIs, I'm assuming that's manual - there's no way to tell the openapi generator that you want to use numpy arrays or whatever...

Yup we can put our language specific files in the generate content to complete a package. At least through that way we won't have to take care of authentication and barebone APIs.

AlexDBlack:
@ShamsUlAzeem Thanks for the answers.
Let's update the ADRs with those clarifications and links (I'd be fine with a Q/A section at the end, so it could be just about copy/paste from here)
Then I'm happy to accept both of the ADRs, unless anyone else wants to provide any feedback (@agibsonccc maybe? Or just tagging for FYI)

ShamsUlAzeem:
@AlexDBlack I've updated the ADRs based on the above discussion and suggestions

ShamsUlAzeem:
@AlexDBlack marked them as accepted

ADRs/0001-Extend_vertx_launcher_for_CLI.md
```diff
@@ -0,0 +1,97 @@
...
+
+Proposed by: Shams Ul Azeem (18-03-2020)
+
+Discussed with: Paul Dubs, Alex Black
+
+## Context
+
+Currently, we have a main class called [KonduitServingMain](https://github.com/KonduitAI/konduit-serving/blob/45af79d15abe4912ccd81e78c9d215306472036e/konduit-serving-core/src/main/java/ai/konduit/serving/configprovider/KonduitServingMain.java) that is the entrypoint for a konduit-serving application to run. The main command line arguments are defined inside [KonduitServingNodeConfigurer](https://github.com/KonduitAI/konduit-serving/blob/e791741b80721980f8b66a35ed42f20b30612d5c/konduit-serving-core/src/main/java/ai/konduit/serving/configprovider/KonduitServingNodeConfigurer.java) class. We also have an additional [Python CLI](https://github.com/KonduitAI/konduit-serving/blob/7965965b58217f2b4d983fd41aaea013264491ee/python/cli.py) that can be just implemented in Java. Vert.x Launcher supports the ability to start, stop and list running verticles out of the box.
+
+## Proposal
```
treo: Also change this to "Decision" when marking an ADR as accepted.

ShamsUlAzeem: Done!

PR #271, opened 2020-03-22
Fix bug where PipelineExecutioner was getting init'ed on every request

EmergentOrder:


EmergentOrder:
Tested with 100,000 requests where before a few thousand would make it crash.

PR #278, opened 2020-03-24
Add classification metrics

agibsonccc:
Adds classification metrics exposure for prometheus.


konduit-serving-api/src/main/java/ai/konduit/serving/config/ServingConfig.java
```diff
@@ -59,6 +60,9 @@
     @Builder.Default
     private boolean createLoggingEndpoints = false;
 
+    @Builder.Default
+    private List<String> classificationLabels = new ArrayList<>(0);
```
ShamsUlAzeem: Is there any way we can make this clearer that this is used for the `CLASSIFICATION` type metrics. (doesn't need to be in this PR). Since we're going to add more similar metrics later so lots of similar options will make it confusing. This could be similar to how Alex suggested of having steps for metrics. Instead of steps we can have a SET of metrics with their corresponding options.

ShamsUlAzeem: Also, maybe we can toggle metrics. It saves a bit of time for tests and starting servers if metrics are turned off.

agibsonccc: @ShamsUlAzeem It's already more or less there. Metric types are configurable so you can already turn them off at the pipeline level. I get what you mean though. That's why I was wondering if we should add a metrics config object concept or something.

ShamsUlAzeem: Yup something like:

metrics:
  - type: CLASSIFICATION (enum)
     labels: 
        - 1
        - 2
     ...
  - type: GPU
     ...

also if the metrics array is empty we'll assume they're turned off. If we want to have a default metrics options then maybe we can have another flag for toggling metrics that's turned off by default since not everyone would want to use metrics especially for first case usages.

konduit-serving-api/src/main/java/ai/konduit/serving/config/TextConfig.java
```diff
@@ -25,6 +31,10 @@
  *
  * @author Alex Black
  */
+@JsonSubTypes({
+        @JsonSubTypes.Type(value = ClassificationMetricsConfig.class, name = "ClassificationMetricsConfig"),
```
ShamsUlAzeem: the name string can be shorted `"CLASSIFICATION"` seems just alright

konduit-serving-api/src/main/java/ai/konduit/serving/config/ServingConfig.java
```diff
@@ -61,7 +62,7 @@
     private boolean createLoggingEndpoints = false;
 
     @Builder.Default
-    private List<String> classificationLabels = new ArrayList<>(0);
+    private List<MetricsConfig> metricsConfigurations = new ArrayList<>(0);
```
ShamsUlAzeem: Awesome!

konduit-serving-api/src/main/java/ai/konduit/serving/config/TextConfig.java
```diff
@@ -33,12 +33,7 @@
  *
  * @author Alex Black
  */
-@JsonSubTypes({
-        @JsonSubTypes.Type(value = RegressionMetricsConfig.class, name = "RegressionMetricsConfig"),
-        @JsonSubTypes.Type(value = NoOpMetricsConfig.class, name = "NoOpMetricsConfig"),
-        @JsonSubTypes.Type(value = ClassificationMetricsConfig.class, name = "ClassificationMetricsConfig"),
-})
-@JsonTypeInfo(use = NAME, include = PROPERTY)
```
agibsonccc: @AlexDBlack @ShamsUlAzeem

Issue #267, opened 2020-03-14
konduit CLI - make it possible to select KS version (future)


AlexDBlack:
Obviously not a priority until we actually _have_ different versions (i.e., after our second release) - but the CLI should allow the user to select which version of KS they want to use.

Why? Suppose we have a breaking change, and the user doesn't want to update all their configurations etc - they just want to keep using the old version.
Currently it's set up to use master branch only.

How? Probably just using github release tags... it should be straightforward to use the github API to detect versions, and then do the appropriate checkout.

Issue #266, opened 2020-03-14
konduit CLI - combine "init" and "build" into one command


AlexDBlack:
Building on Windows 10, in administrator cmd terminal

Attempt 1:
```
konduit build --os windows-x86_64 --spin all                                                                                                                                                        fatal: cannot change to 'C:\Users\Alex\.konduit\konduit-serving': No such file or directory                                                                                                                                    python: can't open file 'C:\Users\Alex\.konduit\konduit-serving\build_jar.py': [Errno 2] No such file or directory
```
It's an administrator cmd, so it _should_ be able to create these directories. I'm guessing it doesn't even check if they exist or not?

After creating directories manually, attempt 2:
```
konduit build --os windows-x86_64 --spin all
fatal: not a git repository (or any of the parent directories): .git
python: can't open file 'C:\Users\Alex\.konduit\konduit-serving\build_jar.py': [Errno 2] No such file or directory
```

Just having `C:\Users\Alex\.konduit\` and not `C:\Users\Alex\.konduit\konduit-serving` doesn't work either.

AlexDBlack:
non-zero return code is ignored: https://github.com/KonduitAI/konduit-serving/blob/master/python/cli.py#L47

AlexDBlack:
Ok, not broken per-se - just bad design / bad UX.
Users are expected to run `konduit init`.
If they, like me, run `konduit build` without running konduit init first - they get the above.

Also, konduit init only does a directory check - if `.konduit\konduit-serving` exists, it assumes it has been cloned before. In my case, the directory was empty, and it failed with the "not a git repository" issue as above.

IMO we should collapse these two commands into one.

Issue #265, opened 2020-03-14
"konduit" CLI - add version option


AlexDBlack:
Most command line tools have a "-version" or "--version" option.
Konduit serving CLI doesn't have this. Simple addition, but important for users.

Issue #242, opened 2020-02-18
Python/Java YAML interop (consolidate to single format)


AlexDBlack:
I'm currently seeing very different YAML produced by Java, vs. the examples/docs.
I'm resonably sure based on the formats that we can't at present create YAML in Java and load it in Python - or vice versa.
If this is the case, it needs to be reproduced in tests and fixed.
For example a series of tests that do: `Java -> yaml -> Python -> yaml -> Java -> yaml` for each type of configuration class.

For details, see:
https://github.com/KonduitAI/konduit-serving/issues/213#issuecomment-587299154

PR #285, opened 2020-04-01
Java CLI

ShamsUlAzeem:
Fixing #231, #242, #265, #266, #267 

Documentation: https://serving.konduit.ai/quickstart/quickstart-cli


konduit-serving-core/src/main/java/ai/konduit/serving/launcher/KonduitServingLauncher.java
```diff
@@ -0,0 +1,153 @@
...
+    /**
+     * Initializes the {@link VertxOptions} for deployment and use in a
+     * {@link Vertx} instance.
+     * The following other initialization also happens:
+     * {@code Vertx Working Directory} gets set (vertx.cwd) and {vertx.caseDirBase)
+     * (vertx.disableFileCPResolving) gets set to true
+     * (vertx.logger-delegate-factory-class-name) gets set to io.vertx.core.logging.SLF4JLogDelegateFactory
+     * The {@link MeterRegistry} field and associated prometheus configuration gets setup
+     * The {@link VertxOptions} event but options also get set
+     */
```
AlexDBlack, line 96: Javadoc describes what happens in static initializer block, not this method

konduit-serving-core/src/main/java/ai/konduit/serving/launcher/KonduitServingLauncher.java
```diff
@@ -0,0 +1,153 @@
...
+            }
+
+            @Override
+            public Verticle createVerticle(String verticleName, ClassLoader classLoader) throws ClassNotFoundException, IllegalAccessException, InstantiationException {
+                String serviceType = verticleName.replace(KONDUIT_PREFIX + ":", "");
+
+                if(services.containsKey(serviceType)) {
+                    return (Verticle) ClassLoader.getSystemClassLoader().loadClass(services.get(serviceType)).newInstance();
+                } else {
+                    log.error(String.format("Invalid service type %s. Possible values are: ", Arrays.asList(services.keySet())));
```
AlexDBlack, line 126: Missing service type, don't need to create list, don't use String.format
```
log.error("Invalid service type \"{}\". Possible values are: {}", serviceType, services.keySet());
```

konduit-serving-core/src/main/java/ai/konduit/serving/launcher/KonduitServingLauncher.java
```diff
@@ -0,0 +1,153 @@
...
+                .unregister("test")
+                .unregister("version")
+                .register(ServeCommand.class, ServeCommand::new)
+                .register(RunCommand.class, RunCommand::new)
+                .register(ListCommand.class, ListCommand::new)
+                .register(StopCommand.class, StopCommand::new)
+                .register(PredictCommand.class, PredictCommand::new)
+                .register(VersionCommand.class, VersionCommand::new)
+                .register(ConfigCommand.class, ConfigCommand::new)
+                .dispatch(args);
```
AlexDBlack, line 151: :+1:

konduit-serving-core/src/main/java/ai/konduit/serving/launcher/command/ConfigCommand.java
```diff
@@ -0,0 +1,263 @@
...
+            if(!outputFile.isFile()) {
+                log.error(String.format("\'%s\' is not a valid file location", outputFile));
+            }
+        } else {
+            try {
+                if(!outputFile.createNewFile()) {
+                    log.error(String.format("\'%s\' is not a valid file location", outputFile));
+                }
+            } catch (Exception exception) {
+                log.error(String.format("Error while creating file: \'%s\'", outputFile), exception);
```
AlexDBlack, line 94: More String.format to remove from logging

konduit-serving-core/src/main/java/ai/konduit/serving/launcher/command/ConfigCommand.java
```diff
@@ -0,0 +1,263 @@
...
+                        ServingConfig.builder()
+                                .build(),
+                        null
+                ).toJson()
+        );
+    }
+
+    private void printOrSave(String output) {
+        if(outputFile == null) {
+            log.info(output);
```
AlexDBlack, line 253: Not sure if it's possible given the system.out to logging redirect, but this might be one of the cases where a System.out.println would be better than a log.info
The reason: Presumably the user will copy/paste this. With log.info, users need to delete the logging prefix part:
i.e., the output will probably look like this:
```
a.k.s.l.c.ConfigCommand - {
a.k.s.l.c.ConfigCommand -   "someField": "someValue"
a.k.s.l.c.ConfigCommand -   "someObject": {
a.k.s.l.c.ConfigCommand -       "otherField": "otherValue"
a.k.s.l.c.ConfigCommand -   }
a.k.s.l.c.ConfigCommand - }
```

konduit-serving-core/src/main/java/ai/konduit/serving/launcher/command/ConfigCommand.java
```diff
@@ -0,0 +1,263 @@
...
+        } else {
+            try {
+                FileUtils.writeStringToFile(outputFile, output, StandardCharsets.UTF_8);
+                log.info("Config file create successfully at {}", outputFile.getAbsolutePath());
+            } catch (IOException exception) {
+                log.error(String.format("Unable to save configuration file to %s", outputFile.getAbsolutePath()), exception);
```
AlexDBlack, line 259: No String.format

konduit-serving-core/src/main/java/ai/konduit/serving/launcher/command/ServeCommand.java
```diff
@@ -0,0 +1,236 @@
...
+
+        // Add id - it's important as it's the application mark.
+        args.add("-Dserving.id=" + getId());
+        return args;
+    }
+
+    private String getId() {
+        if (id == null) {
+            id = UUID.randomUUID().toString();
+        }
```
AlexDBlack, line 204: Let's do something a little more user friendly than full UUIDs for default IDs, which end up looking like this:
```
20767bd7-0ef1-4164-80a4-b6391c727e69
```

Some options:
1. First 8 characters of UUID only
2. Millisecond timestamp (1585967431586 etc)
3. Human readable timestamp: YYYYMMDD-HHMMSS.sss
4. Some sort of "two word" name generator system like Docker has - https://github.com/moby/moby/blob/master/pkg/namesgenerator/names-generator.go

Maybe 1, but we revisit 4 later.

konduit-serving-core/src/main/java/ai/konduit/serving/launcher/command/VersionCommand.java
```diff
@@ -0,0 +1,66 @@
...
+import io.vertx.core.spi.launcher.DefaultCommand;
+import lombok.extern.slf4j.Slf4j;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Scanner;
+
+@Name(value = "version", priority = 1)
+@Summary("Displays konduit-serving version.")
+@Description("Prints the vert.x core version used by the application.")
```
AlexDBlack, line 34: `Prints the vert.x core version` -> `Prints the konduit-serving version`?

konduit-serving-core/src/main/java/ai/konduit/serving/launcher/command/ConfigCommand.java
ShamsUlAzeem: Yeah took care of that programmatically. Trying to filter out only relevant information here. Still working on it. I realised there's quite a bit of work needed on spitting out useful logs that's only relevant to a user. Kind of like a simplified way of presenting an error and a possible solution if there's a mistake on the user's end. I'm normally thinking of keeping all the logs in a `command.log` which can be useful for reporting errors.

konduit-serving-core/src/main/java/ai/konduit/serving/launcher/command/ConfigCommand.java
ShamsUlAzeem: Sorry about that. Sometimes it doesn't work for me so I usually go with String.format. Maybe I have to explicitly set `{0}`, `{1}` inside the log string formats.

konduit-serving-core/src/main/java/ai/konduit/serving/launcher/command/ServeCommand.java
ShamsUlAzeem: Yup I'm gonna use JavaFaker at the end over here: https://github.com/DiUS/java-faker. Seems like a good solution for generation random names that's not already taken.

konduit-serving-core/src/main/java/ai/konduit/serving/launcher/command/VersionCommand.java
ShamsUlAzeem: Ah thanks!

konduit-serving-api/src/main/java/ai/konduit/serving/config/ServingConfig.java
```diff
@@ -53,7 +53,7 @@
     private Output.DataFormat outputDataFormat = Output.DataFormat.JSON;
 
     @Builder.Default
-    private String uploadsDirectory = "file-uploads/";
+    private String uploadsDirectory = System.getProperty("java.io.tmpdir");
```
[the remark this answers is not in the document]
ShamsUlAzeem: Sure

konduit-serving-core/src/main/java/ai/konduit/serving/deploy/DeployKonduitServing.java
```diff
@@ -0,0 +1,107 @@
...
+                if (eventHandler != null) {
+                    VertxImpl vertxImpl = (VertxImpl) vertx;
+                    DeploymentOptions inferenceDeploymentOptions = vertxImpl.getDeployment(handler.result()).deploymentOptions();
+
+                    try {
+                        InferenceConfiguration inferenceConfiguration = ObjectMappers.fromJson(inferenceDeploymentOptions.getConfig().encode(), InferenceConfiguration.class);
+                        eventHandler.handle(Future.succeededFuture(inferenceConfiguration));
+                    } catch (Exception exception){
+                        log.debug("Unable to parse json configuration into an InferenceConfiguration object. " +
+                                "This can be ignored if the verticle isn't an InferenceVerticle.", exception);
```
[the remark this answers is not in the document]
ShamsUlAzeem: There's one more verticle that we might use in the future. https://github.com/KonduitAI/konduit-serving/blob/45af79d15abe4912ccd81e78c9d215306472036e/konduit-serving-converter/src/main/java/ai/konduit/serving/input/converter/ConverterVerticle.java

konduit-serving-core/src/main/java/ai/konduit/serving/launcher/KonduitServingLauncher.java
ShamsUlAzeem: Oh my bad...

konduit-serving-core/src/main/java/ai/konduit/serving/deploy/DeployKonduitServing.java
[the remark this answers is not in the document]
AlexDBlack: OK, fair. Maybe add a brief comment on that here...

konduit-serving-core/src/main/java/ai/konduit/serving/launcher/command/ConfigCommand.java
AlexDBlack: hm... so you're suggesting we split into say "user logs" and "debug/system logs"
Basically the same thing but:
* User logs go to console only, info level and higher, _maybe_ with different formatting (no class prefix)?
* System logs go to file, are debug level and higher?

Something like that would be good for both debugging and usability... let's open an issue to discuss further.
And yes, we no doubt need to do a pass on log levels to ensure good UX for users

konduit-serving-core/src/main/java/ai/konduit/serving/launcher/command/ConfigCommand.java
AlexDBlack: I've never had to use anything other than `{}` in logging strings in all of DL4J and ND4J for logging - the only issue is that you don't need a {} for exceptions (it'll append that at the end with the stack trace)

konduit-serving-core/src/main/java/ai/konduit/serving/launcher/command/ServeCommand.java
ShamsUlAzeem: Using the first 8 characters of UUID for now

konduit-serving-core/pom.xml
```diff
@@ -36,6 +36,27 @@
...
+                        <goals>
+                            <goal>revision</goal>
+                        </goals>
+                        <phase>generate-resources</phase>
+                    </execution>
+                </executions>
+                <configuration>
+                    <generateGitPropertiesFile>true</generateGitPropertiesFile>
+                    <generateGitPropertiesFilename>
+                        src/main/resources/META-INF/git.properties
```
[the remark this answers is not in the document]
ShamsUlAzeem: Didn't use the property but I've updated the name.

konduit-serving-core/src/main/java/ai/konduit/serving/launcher/command/KonduitRunCommand.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    private String getServerId() {
+        String[] commandSplits = CommandLineUtils.getCommand().split(" ");
+        String lastSegment = commandSplits[commandSplits.length - 1];
+        if(lastSegment.contains("serving.id")) {
+            return lastSegment.replace("-Dserving.id=", "").trim();
+        } else {
+            return null;
+        }
```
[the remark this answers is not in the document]
ShamsUlAzeem: This makes sure that it's at the end of the command line arguments: https://github.com/KonduitAI/konduit-serving/blob/03567a83bce210f2457ae22beb493b876ed319d2/konduit-serving-core/src/main/java/ai/konduit/serving/launcher/command/ServeCommand.java#L297

Also, since the command parser for the `serve` command would ignore spaces while taking arguments so we won't have to worry about that here.

ShamsUlAzeem:
@AlexDBlack ready for review now. Couldn't get the test coverage to 80% but this much should be fine in my opinion. Let me know if I should change something. 

About the python CLI. I've added the logic for downloading the jar file if it's not present and then channeling the arguments to the main java CLI but we still need to figure out hosting the relevant jar files. Can we upload jar files into pip packages?

PR #287, opened 2020-04-02
[OLD API] (WIP) Extract out onnx

agibsonccc:
Moves onnx to its own module (Same logic as other modules, keeps core lean and allows us to run on arm)


agibsonccc:
@sshepel I believe you'll need to update your build scripts to reflect this.

agibsonccc:
@sshepel I removed tfjava as a default module too, so you may need to add that as well. Those are just profile flags.

sshepel:
`onnx` profile has been added to mvn call in CI

ShamsUlAzeem:
Closed due to becoming old api

PR #288, opened 2020-04-05
Change classification metrics to reset upon export

agibsonccc:
Classification counts are inflated unless they are reset upon export.  Metrics for classification are now reset, and it changes the metrics export to use a gauge rather than a counter so value scan go up and down.

### DIFF
--- a/konduit-serving-core/src/main/java/ai/konduit/serving/metrics/RegressionMetrics.java
+++ b/konduit-serving-core/src/main/java/ai/konduit/serving/metrics/RegressionMetrics.java
@@ -28,6 +28,7 @@ import ai.konduit.serving.config.metrics.MetricsRenderer;
 import ai.konduit.serving.config.metrics.impl.RegressionMetricsConfig;
 import ai.konduit.serving.util.MetricRenderUtils;
 import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.ImmutableTag;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Tag;
 import lombok.Getter;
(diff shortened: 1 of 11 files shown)


konduit-serving-core/src/main/java/ai/konduit/serving/metrics/RegressionMetrics.java
```diff
@@ -145,11 +148,11 @@ public MetricsConfig config() {
...
     public void updateMetrics(Object... args) {
         if(args instanceof Record[]) {
-            Record[] records = (Record[]) args[0];
+            Record[] records = (Record[]) args;
             incrementClassificationCounters(records);
         }
         else if(args instanceof INDArray[]) {
-            INDArray[] output = (INDArray[]) args[0];
+            INDArray[] output = (INDArray[]) args;
             incrementClassificationCounters(output);
```
ShamsUlAzeem, line 157: Maybe write it as you've done for the classification metrics:
```
        if(args[0] instanceof Record) {
            Record records = (Record) args[0];
            incrementClassificationCounters(new Record[]{records});
        }
        else if(args[0] instanceof Record[]) {
            Record[] records = (Record[]) args[0];
            incrementClassificationCounters(records);
        }
        else if(args[0] instanceof INDArray) {
            INDArray output = (INDArray) args[0];
            incrementClassificationCounters(new INDArray[] {output});
        }
        else if(args[0] instanceof INDArray[]) {
            INDArray[] output = (INDArray[]) args[0];
            incrementClassificationCounters(output);
        }
```

so that the records can be passed in as both `updateMetrics(new INDArray[] {...})` or `updateMetrics(INDArray, INDArray)`

PR #291, opened 2020-04-16
add missing modules to execute "konduit serve --config config.yml"

kmotohas:
added the following lines to `load.py` to avoid `NameError: name '...' is not defined` when `konduit serve --config ...` is executed.

```
from .base_inference import ServingConfig
from .base_inference import PythonConfig
from .base_inference import PythonStep
```

### DIFF
--- a/python/konduit/load.py
+++ b/python/konduit/load.py
@@ -6,6 +6,9 @@ import yaml
 from .client import Client
 from .server import Server
 from .utils import to_unix_path, update_dict_with_unix_paths
+from .base_inference import ServingConfig
+from .base_inference import PythonConfig
+from .base_inference import PythonStep
 
 KONDUIT_BASE_DIR = os.getcwd()
 KONDUIT_PID_STORAGE = os.path.join(KONDUIT_BASE_DIR, "pid.json")


kmotohas:
by the way, `ServingConfig` in `base_inference.py` does not have `input_data_format` attribute.
Is this intended?
I needed to comment out it from yaml file.
```
serving:
  http_port: 1337
#  input_data_format: NUMPY
  output_data_format: NUMPY
  log_timings: True
  extra_start_args: -Xmx8g
steps:
  python_step:
    type: PYTHON
    python_path: .
    python_code_path: ./simple.py
    python_inputs:
      first: NDARRAY
    python_outputs:
      second: NDARRAY
client:
    port: 1337
```

AlexDBlack:
The Java serving configuration class similarly does not have an input data format configuration, only an output data format configuration - https://github.com/KonduitAI/konduit-serving/blob/master/konduit-serving-api/src/main/java/ai/konduit/serving/config/ServingConfig.java

I believe that this is due to the endpoints encoding the input type - `/:predictionType/:inputDataFormat`

So I think this is a documentation issue - i.e., this page shouldn't show an input_data_format flag - https://serving.konduit.ai/yaml-configurations#yaml-components

cc @ShamsUlAzeem

AlexDBlack:
Also CI here is failing on ONNX tests
https://github.com/KonduitAI/konduit-serving/issues/295
This failure is unrelated to (not caused by) your changes.

I've restarted CI

ShamsUlAzeem:
The Onnx tests started failing in MacOS I can reproduce it there. Works fine on windows locally. 

the input_data_format was removed due to redundancy in the serving configuration. I'll update the documentation. Let's merge this one.

Issue #298, opened 2020-04-29
Implement Data/API refactoring ADR (Implementation Plan)


AlexDBlack:
This issue is about how we will proceed with implementing this ADR:
https://github.com/KonduitAI/konduit-serving/blob/master/ADRs/0003-Pipeline_API_Rewrite.md

For the _what_ of the implementation - see the ADR. This issue focuses exclusively on the _how_ of actually executing the implementation.

---
---

Implementing the Pipeline API Rewrite ADR will take a significant amount of time and will impact virtually all aspects of Konduit Serving.

During the buildout/implementation, we want to avoid the following:
* Breaking Konduit Serving on an implementation branch and then not having anything work (or be mergeable) until everything is entirely complete
* Having a huge PR that can't be reviewed effectively due to its large size
* Merge conflicts if master branch is continued to be updated in parallel
* Difficulties in parallelizing the build-out between different engineers due to a monolithic, single-branch rewrite approach.

**The plan**: we will have the new API and old API functionality co-exist in parallel in the master branch of the repository during the duration of ADR implementation.
The modifications for the new API will be worked on without breaking the existing functionality/tests, with code duplication in additional modules/packages when necessary. The new functionality merged frequently, in stages, even if it's not "ready" - other than for new tests, the new API functionality will not be used by Konduit Serving users until the refactoring is complete.
Once the new API functionality is complete and is adequately tested, we can delete the old classes and make it "live".

The main downside of this approach is code duplication.

The actual build-out will proceed in phases, and can be performed in parallel:
Step 1 -> 3 -> (4, 5, 6) -> (7,8)
Step 2 can be done at any point

**Step 1**: Add Data interface, JData and Protobuf implementations
Currently underway.

~~**Step 2**: Deprecate API classes/interfaces, etc to be removed in the new buildout~~
~~outputAdapter, MultiOutputAdapter, Batchoutput, Input, Output, SchemaType, BatchOutput, TensorDataType~~
Complete: https://github.com/KonduitAI/konduit-serving/pull/299

~~**Step 3**: create new pipeline API module; duplicate and update the API for main interfaces
Specifically, duplicate the core interfaces in konduit-serving-api~~
Complete: https://github.com/KonduitAI/konduit-serving/pull/299

We will add a new module: konduit-serving-pipeline
In it, put a new copy of the following API classes/interfaces in ai.konduit.serving.api: PipelineStep, PipelineStepRunner, and Pipeline
[...]

PR #299, opened 2020-04-29
Add new API + konduit-serving-pipeline module

AlexDBlack:
Implements steps 2 and 3 as described here: https://github.com/KonduitAI/konduit-serving/issues/298

Feedback on design here would be appreciated, though of course we can make changes after it is merged as needed.


AlexDBlack:
Merging this now. Comments/feedback are still welcome.

PR #321, opened 2020-05-15
[OLD API] Python CLI module updates (CUDA VERSION and creating/uploading artifacts) [WIP]

ShamsUlAzeem:
Adds cuda versions to the GPU builds and automated scripts for creating and uploading binary artifacts to release tags.

Also, this adds the ability to pass in extra arguments to the python build_jar.py command. 

Requires python package update: `pip install --upgrade --force-reinstall --no-deps konduit`


AlexDBlack:
In principle - this looks good. In practice - there are some challenges to 

There's 2 problems / concerns here:
(a) the javacpp version that CUDA 10.0 might be pulling in
we need to carefully check that it's not pulling in both javacpp 1.5 and 1.5.2, otherwise we're likely to run into problems.

(b) the fact that TF doesn't release with multiple CUDA versions.
For example TF 2.2 and 2.1 is CUDA 10.1 only, unless you build from source
and TF 1.15.x is CUDA 10.0 only

Similarly, javacpp tensorflow-platform-gpu also doesn't have a CUDA version, i.e., it supports only the same CUDA version as TF itself

AlexDBlack:
I was thinking that a solution may be to simply to standardize on the TF CUDA version (10.0 due to TF 1.15 dependency), bundle the CUDA version we need using org.bytedeco:cuda-platform-redist
https://github.com/bytedeco/javacpp-presets/tree/master/cuda

The problem is: there's currently no cuda-platform-redist for 10.0, only for 10.1 and 10.2
https://repo1.maven.org/maven2/org/bytedeco/cuda-platform-redist/

ShamsUlAzeem:
So, due to enforcement rules only one javacpp version is going to be included in the uberjar. Although there will be a warning for cuda-platform javacpp version not being equal to the included javacpp version but that's not going to make much difference. 

For tensorflow, I'll include an exception while starting the konduit server that the cuda version is unsupported. Other than that there are certain graph-def limitations for backward compatibility between 1.15 and 2.x. If it's possible I'll add a check for that so the model files are compatible with the included konduit serving tensorflow version. I'll point to a documentation page in the error message for a way to compile for 1.15.

For CUDA versions, this is what I'm thinking for options:
- Cuda 10.0 ==>  Local CUDA Installation OR Docker build
- Cuda 10.1 ==>  Local CUDA Installation OR CUDA redist OR Docker build
- Cuda 10.2 ==>  Local CUDA Installation OR CUDA redist OR Docker build

pom.xml
```diff
@@ -738,6 +729,11 @@
             <id>konduit-packages</id>
             <url>https://packages.konduit.ai/repository/maven-public/</url>
         </repository>
+        <repository>
+            <id>sonatype-nexus-staging</id>
+            <name>Sonatype Nexus Staging</name>
+            <url>https://oss.sonatype.org/content/repositories/orgdeeplearning4j-1115/</url>
+        </repository>
```
AlexDBlack, line 736: Remove

ShamsUlAzeem:
Closing as it's  part of old APIs

Issue #322, opened 2020-05-15
Implement Profiler functionality


AlexDBlack:
Currently we have a placeholder Profiler interface
https://github.com/KonduitAI/konduit-serving/blob/master/konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/api/context/Profiler.java

We will need to make this actually work and be useful for debugging and understanding performance of our pipelines, on a per-step basis.

I suggest we have two implementations of the Profiler interface:
1. One that is just a no-op - i.e., for when profiler is disabled
2. One that asynchronously writes profiler data to disk in Chrome Profiler (JSON) format. i.e., file writing is not done in the main thread.

We use chrome profiler format JSON format here:
https://github.com/eclipse/deeplearning4j/blob/master/nd4j/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/autodiff/listeners/profiler/ProfilingListener.java#L45
Note there's both array and object format - let's use Array format - see this for more details: https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/edit
Time unit should be microseconds. We'll use duration events here. PID and TID should be logged. Category should be the name of the pipeline step class (i.e., "ImageToNDArrayStep" etc).

We will also add a ProfilerConfig class - so we can do something like PipelineExecutor.profile(new ProfilerConfig(...))
ProfilerConfig should allow the user to specify the output file location, and file size split in bytes (i.e., if profile file is later than X, create a new file - kind of like log rotation)

But unlike in SameDiff, we won't have a warmup setting or anything.
Now, when the profiler is enabled:
- we should automatically log the start and end of each PipelineStep exec call, within SequencePipelineExecutor.exec(Data) method
- For "within step" profiling, we should also implement the behaviour described in Profiler interface javadoc: i.e., check that if the user calls eventStart, but not eventEnd, we automatically call eventEnd after PipelineStepRunner.exec(...) returns

We won't make any changes to GraphPipelineExecutor.exec(Data) for now, as there will be some major changes there soon (see open ADR for graph pipelines)

AlexDBlack:
https://github.com/KonduitAI/konduit-serving/pull/324

PR #324, opened 2020-05-18
Pipelines profiler

alexanderst:


konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/api/context/TraceEvent.java
```diff
@@ -0,0 +1,19 @@
...
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+public class TraceEvent {
+    public static enum EventType {
+        START, END;
+    }
+    private String name;
+    private long timeStamp;
+    EventType type;
```
AlexDBlack, line 18: If we're using TraceEvent to directly write JSON, the format (and hence fields names) need to match the chrome tracing format: "name", "ts", "ph".
As it stands, the JSON produced by serializing this class won't be readable.

https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/edit

We also need process ID, thread ID, and category.
See also TraceEvent class in ND4J (for SameDiff).

Alternatively for JSON serialization we can use a  custom JSON serializer - see DataJsonSerializer and the JsonSerialize/Deserialize on Data interface for an example on this. But simply naming the fields to match the required JSON format is easier IMO.

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/api/pipeline/PipelineExecutor.java
```diff
@@ -76,4 +78,9 @@ default void close() {
      * Get the logger used by this PipelineExecutor
      */
     Logger getLogger();
+
+    /**
+     * Profiling
+     */
+    void profilerConfig(Profiler profiler);
```
AlexDBlack, line 85: `void profilerConfig(ProfilerConfig profiler);`

konduit-serving-pipeline/src/test/java/ai/konduit/serving/pipeline/impl/pipeline/PipelineProfilerTest.java
```diff
@@ -0,0 +1,54 @@
...
+        pe.profilerConfig(profiler);
+        pe.exec(d);
+
+        StringBuilder sb = new StringBuilder();
+        try (BufferedReader br = new BufferedReader(new FileReader(logFile))) {
+            String line;
+            while ((line = br.readLine()) != null)
+                sb.append(line).append("\n");
+        }
+        System.out.println(sb.toString());
```
AlexDBlack, line 52: Use one of these methods instead if you need to read file contents.
```
import org.apache.commons.io.FileUtils;
String fileContents = FileUtils.readFileToString(logFile, StandardCharsets.UTF_8);
List<String> lines = FileUtils.readLines(logFile, StandardCharsets.UTF_8);
```

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/impl/pipeline/SequencePipelineExecutor.java
```diff
@@ -66,7 +69,13 @@ public Data exec(Data data) {
         Context ctx = null; //TODO
         Data current = data;
         for(PipelineStepRunner psr : runners){
+            if (profiler.profilerEnabled()) {
+                profiler.eventStart(psr.toString());
+            }
             current = psr.exec(ctx, current);
+            if (profiler.profilerEnabled()) {
+                profiler.eventEnd(psr.toString());
```
AlexDBlack, line 77: For the profile name:
(a) Add a PipelineStep.name() method:
```
    default String name(){
        return getClass().getSimpleName();
    }
```

(b) use `profiler.eventStart(psr.getPipelineStep().name());` instead

alexanderst:
> Also copyright headers needed.
> You can configure IntelliJ to automatically insert them for new files FYI
> File -> Settings -> Editor -> Copyright -> Copyright profiles -> (create profile and add text)
> here's profile text I use: https://gist.github.com/AlexDBlack/7593aa00bb69bb3537151d9baa88d619
> Then File -> Settings -> Editor -> (select profile in "Default project copyright")
> 
> Don't forget the log rotation / file splitting and the "eventStart called but no eventEnd called"
> #322

Is it ok to remove old file for rotation?

AlexDBlack:
> Is it ok to remove old file for rotation?

No, let's keep the old file for now, and just append an index to the filename - for example "myFile.json", "myFile_1.json", "myFile_2.json" etc

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/api/context/PipelineProfiler.java
```diff
@@ -0,0 +1,204 @@
...
+                .ts(endTime)
+                .ph(TraceEvent.EventType.END)
+                .pid(this.pid)
+                .tid(this.tid)
+                .build();
+
+        writeQueue.add(event);
+    }
+
+    public TraceEvent[] readEvents(File file) throws IOException {
```
AlexDBlack, line 194: This can be public static

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/api/context/PipelineProfiler.java
```diff
@@ -0,0 +1,204 @@
...
+            int num = 1;
+            if (parts.length > 1) {
+                num = Integer.parseInt(parts[1]) + 1;
+            }*/
+            Path backup = Paths.get(currentFile.getParent() + FileSystems.getDefault().getSeparator() +
+                    baseName + "_" + String.valueOf(System.currentTimeMillis()) + ".json");
+            Files.move(currentFile, backup);
+            Files.createFile(currentFile);
+        }
+    }
```
AlexDBlack, line 91: We should only call fileSizeGuard() from the writing thread, to avoid race conditions.
i.e., it might (try to) move the file right when the thread is in the middle of writing. Plus given we're using a buffered writer, we should be closing (or at very least flushing) the writer before we do anything to the file.

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/api/context/Profiler.java
```diff
@@ -65,12 +67,12 @@
...
      */
-    void eventStart(String key);
+    void eventStart(String key) throws IOException;
 
     /**
      * End the timer for the event with the specified key.
      * Should be called some time after {@link #eventStart(String)}
      */
-    void eventEnd(String key);
+    void eventEnd(String key) throws IOException;
```
AlexDBlack, line 76: Remove these checked exceptions 
We don't want users/devs to have to add a try/catch every single time they call a profiler method.

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/api/context/ProfilerConfig.java
```diff
@@ -0,0 +1,16 @@
+package ai.konduit.serving.pipeline.api.context;
```
AlexDBlack, line 1: Copyright header.

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/api/context/TraceEvent.java
```diff
@@ -0,0 +1,41 @@
...
+    public static enum EventType {
+        START, END;
+    }
+    private String name;
+    private List<String> cat;
+    private long ts;
+    private long pid;
+    private long tid;
+    EventType ph;
+}
```
AlexDBlack, line 41: Two more small change needed to comply with the Chrome profiler trace format:
(a) cat - should be a String (comma separated) not a List<String> - that will produce JSON like `["a", "b"]` not `"a,b"`
(b) EventType enum - needs to be "B" and "E" - (see table at end of page 4 - https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/edit )

Also access modifier - `private EventType ph`

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/api/context/PipelineProfiler.java
```diff
@@ -0,0 +1,213 @@
...
+                } catch (InterruptedException e) {
+                    throw new RuntimeException(e);
+                }
+            }
+            try {
+                writer.flush();
+            } catch (IOException e) {
+                throw new RuntimeException(e);
+            }
+        }
```
AlexDBlack, line 187: This can't be in the logging method. The whole point of having a separate thread for file is to make sure we don't add _any_ overhead to the main thread. If you properly flush and close the writer in fileSizeGuard() this shouldn't be necessary either.

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/api/context/PipelineProfiler.java
```diff
@@ -0,0 +1,213 @@
...
+            Path nextFile = Paths.get(currentLog.getParent() + FileSystems.getDefault().getSeparator() +
+                    counted + ".json");
+            Files.createFile(nextFile);
+            try {
+                this.writer = new BufferedWriter(new FileWriter(nextFile.toString(), true));
+                this.writer.write("[");     //JSON array open (array close is optional for Chrome profiler format)
+            } catch (IOException e) {
+                throw new RuntimeException(e);
+            }
+            this.currentLog = nextFile;
```
AlexDBlack, line 98: You should be flushing and closing the existing writer before creating the next one.

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/api/pipeline/PipelineExecutor.java
AlexDBlack: Still needs to be changed. User provides the config, we instantiate the Profiler instance internally.

alexanderst: Where should we create Profiler object in snippet below? It depends on profilerConfig, so it must be created in pe.exec() method. Is it always safe? I'd move
 creation of profiler to PipelineExecutor constructor.

```
PipelineExecutor pe = p.executor();
pe.profilerConfig(new ProfilerConfig());
pe.exec(d);
```

PR #317, opened 2020-05-12
Vertx module implementation

ShamsUlAzeem:
This PR continues `Step 7` from the issue defined here: #298 

Adds `konduit-serving-vertx` module and related protocol implementations. The protocols are implemented as child modules inside `konduit-serving-vertx-protocols` module. `konduit-serving-vertx` module specifies the base classes for deploying inference verticles where as the protocols are registered on runtime based on what's included in the classpath. The final structure will look like this:

- konduit-serving-vertx
- konduit-serving-vertx-protocols
  - konduit-serving-http
  - konduit-serving-mqtt
  - konduit-serving-grpc

Modules `konduit-serving-mqtt` and `konduit-serving-grpc` will be added in a separate PR.

PR #325, opened 2020-05-19
GRPC and MQTT modules implementation

ShamsUlAzeem:
Continuation of #317 for the implementation of `konduit-serving-grpc` and `konduit-serving-mqtt` modules.


konduit-serving-vertx-protocols/konduit-serving-grpc/pom.xml
```diff
@@ -0,0 +1,77 @@
...
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.xolstice.maven.plugins</groupId>
+                <artifactId>protobuf-maven-plugin</artifactId>
+                <version>0.5.0</version>
+                <configuration>
+                    <protocArtifact>com.google.protobuf:protoc:3.2.0:exe:${os.detected.classifier}</protocArtifact>
```
AlexDBlack, line 47: Also we've got protoc-jar-maven-plugin in konduit-serving-pipeline... I'm guessing we can't use that because we need the io.vertx:protoc-gen-grpc-java plugin?

If we can't use the other pligun, there's still Protobuf 3.2.0 here vs. 3.8.0 used for generation in konduit-serving-pipeline.

ShamsUlAzeem: I've used the `protoc-jar-maven-plugin` with the vertx plugin artifacts but the output wasn't what I needed so I'm using this plugin with the updated protobuf version

ShamsUlAzeem:
@AlexDBlack changed the base to master as well. I didn't sync it with master before. That's why it was showing extra commit lines.

PR #343, opened 2020-05-27
Switch TF to javacpp presets (and 1.15.0 -> 1.15.2) [WIP]

AlexDBlack:
This works, but:
* TensorFlowUtil.ensureNativeLibrariesLoaded(); - maybe a better approach is available? (`@Platform` annotation maybe, but we need to ensure the annotated class is guaranteed to be loaded before anything that uses TF...
* mkl-dnn (necessary for x86) should be in a profile to enable to build on non-x86 systems, but I couldn't get the profile to activate during tests based on os.arch property (which is definitely set correctly)

### DIFF
--- a/konduit-serving-models/konduit-serving-tensorflow/pom.xml
+++ b/konduit-serving-models/konduit-serving-tensorflow/pom.xml
@@ -31,9 +31,34 @@
 
     <dependencies>
         <dependency>
-            <groupId>org.tensorflow</groupId>
+            <groupId>org.bytedeco</groupId>
             <artifactId>tensorflow</artifactId>
-            <version>1.15.0</version>
+            <version>1.15.2-1.5.3</version>
+        </dependency>
+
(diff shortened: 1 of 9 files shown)


saudet:
> * TensorFlowUtil.ensureNativeLibrariesLoaded(); - maybe a better approach is available? (`@Platform` annotation maybe, but we need to ensure the annotated class is guaranteed to be loaded before anything that uses TF...

Call `Loader.load(org.bytedeco.tensorflow.presets.tensorflow.class)`, that's what it's for:
https://github.com/bytedeco/javacpp-presets/tree/master/tensorflow#documentation

> * mkl-dnn (necessary for x86) should be in a profile to enable to build on non-x86 systems, but I couldn't get the profile to activate during tests based on os.arch property (which is definitely set correctly)

Like for everything else, use `tensorflow-platform`, not just `tensorflow`.

PR #348, opened 2020-06-02
Bounding Box Filter + test

atuzhykov:

### DIFF
--- a/konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/impl/step/bbox/filter/BoundingBoxFilterStepRunner.java
+++ b/konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/impl/step/bbox/filter/BoundingBoxFilterStepRunner.java
@@ -1,0 +1,58 @@
+package ai.konduit.serving.pipeline.impl.step.bbox.filter;
+
+import ai.konduit.serving.pipeline.api.context.Context;
+import ai.konduit.serving.pipeline.api.data.BoundingBox;
+import ai.konduit.serving.pipeline.api.data.Data;
+import ai.konduit.serving.pipeline.api.data.ValueType;
+import ai.konduit.serving.pipeline.api.step.PipelineStep;
+import ai.konduit.serving.pipeline.api.step.PipelineStepRunner;
+import lombok.AllArgsConstructor;
+
+import java.util.List;
+
+@AllArgsConstructor
+public class BoundingBoxFilterStepRunner implements PipelineStepRunner {
+
+    protected final BoundingBoxFilterStep step;
+
+    @Override
+    public void close() {
+
+    }
+
+    @Override
+    public PipelineStep getPipelineStep() {
+        return step;
+    }
+
+
+    @Override
+    public Data exec(Context ctx, Data data) {
+
+        String[] classesToKeep = step.classesToKeep();
+        System.out.println(data);
+        List<BoundingBox> boundingBoxes = data.getListBoundingBox("img_bbox");
+
+//       bbox has .label as null. why?
+//        for (BoundingBox bbox : boundingBoxes
+//        ) {
+//            for (String classToKeep : classesToKeep) {
+//                if (!bbox.label().startsWith(classToKeep)) {
+//                    boundingBoxes.remove(bbox);
+//                }
+//            }
+//        }
+
+
+        String outName = step.outputName();
+        if (outName == null)
+            outName = BoundingBoxFilterStep.DEFAULT_OUTPUT_NAME;
+
+        Data d = Data.singletonList(outName, boundingBoxes, ValueType.BOUNDING_BOX);
+
+
+        return d;
+    }
+
+
+}
--- a/konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/impl/step/bbox/filter/BoundingBoxFilterStepRunner.java
+++ b/konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/impl/step/bbox/filter/BoundingBoxFilterStepRunner.java
@@ -23,16 +23,22 @@ import ai.konduit.serving.pipeline.api.data.Data;
 import ai.konduit.serving.pipeline.api.data.ValueType;
 import ai.konduit.serving.pipeline.api.step.PipelineStep;
 import ai.konduit.serving.pipeline.api.step.PipelineStepRunner;
-import lombok.AllArgsConstructor;
+import ai.konduit.serving.pipeline.util.DataUtils;
+import org.apache.commons.lang3.ArrayUtils;
+import org.nd4j.shade.guava.base.Preconditions;
 
 import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
 
-@AllArgsConstructor
 public class BoundingBoxFilterStepRunner implements PipelineStepRunner {
 
     protected final BoundingBoxFilterStep step;
+
+    public BoundingBoxFilterStepRunner(BoundingBoxFilterStep step) {
+        this.step = step;
+        Preconditions.checkArgument(!ArrayUtils.isEmpty(this.step.classesToKeep),"Seems you forgets to set the classes to keep.");
+    };
 
     @Override
     public void close() {
@@ -51,6 +57,11 @@ public class BoundingBoxFilterStepRunner implements PipelineStepRunner {
         String prob = "detection_scores";
         String labels = "detection_classes";
 
+        String inputName = step.inputName();
+        if(inputName == null){
+            String err = "No input name was set in the BoundingBoxFilterStep configuration and input name could not be guessed based on type";
+            DataUtils.inferField(data, ValueType.BOUNDING_BOX, true, err + " (multiple keys)", err + " (no List<BoundingBox> values)");
+        }
         String[] classesToKeep = step.classesToKeep();
         List<BoundingBox> boundingBoxes = data
                 .getListBoundingBox(step.inputName)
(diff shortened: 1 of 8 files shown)


konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/impl/step/bbox/filter/BoundingBoxFilterStepRunner.java
```diff
@@ -0,0 +1,53 @@
...
+                .stream()
+                .filter(i -> Arrays.stream(classesToKeep).anyMatch(i.label()::equals))
+                .collect(Collectors.toList());
+
+
+        String outName = step.outputName();
+        if (outName == null)
+            outName = BoundingBoxFilterStep.DEFAULT_OUTPUT_NAME;
+
+        Data d = Data.singletonList(outName, boundingBoxes, ValueType.BOUNDING_BOX);
```
AlexDBlack, line 46: We should be keeping all of the original values. Right now, you're throwing away everything except the bounding box list.
This can be done using Data.copyFrom for all keys except the input one

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/impl/step/bbox/filter/BoundingBoxFilterStepRunner.java
```diff
@@ -0,0 +1,53 @@
...
+        return step;
+    }
+
+
+    @Override
+    public Data exec(Context ctx, Data data) {
+
+        String[] classesToKeep = step.classesToKeep();
+        List<BoundingBox> boundingBoxes = data
+                .getListBoundingBox("img_bbox")
```
AlexDBlack, line 36: Don't hardcode the key, this should be specified in the configuration (BoundingBoxFilterStep)

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/impl/step/ml/ssd/SSDToBoundingBoxRunner.java
```diff
@@ -25,15 +25,19 @@
...
 
 @AllArgsConstructor
 public class SSDToBoundingBoxRunner implements PipelineStepRunner {
 
     protected final SSDToBoundingBoxStep step;
+    protected final String[] COCO_CLASSES_LABELS = new String[]{"person", "bicycle", "car", "motorcycle", "airplane", "bus", "train", "truck", "boat", "traffic light", "fire hydrant", "street sign", "stop sign", "parking meter", "bench", "bird", "cat", "dog", "horse", "sheep", "cow", "elephant", "bear", "zebra", "giraffe", "hat", "backpack", "umbrella", "shoe", "eye glasses", "handbag", "tie", "suitcase", "frisbee", "skis", "snowboard", "sports ball", "kite", "baseball bat", "baseball glove", "skateboard", "surfboard", "tennis racket", "bottle", "plate", "wine glass", "cup", "fork", "knife", "spoon", "bowl", "banana", "apple", "sandwich", "orange", "broccoli", "carrot", "hot dog", "pizza", "donut", "cake", "chair", "couch", "potted plant", "bed", "mirror", "dining table", "window", "desk", "toilet", "door", "tv", "laptop", "mouse", "remote", "keyboard", "cell phone", "microwave", "oven", "toaster", "sink", "refrigerator", "blender", "book", "clock", "vase", "scissors", "teddy bear", "hair drier", "toothbrush", "hair brush"};
```
AlexDBlack, line 39: Two things to change here
(a) the labels should come from the configuration (i.e., something the user sets in SSDToBoundingBoxStep) - and they should be optinal. If the user doesn't provide a list of class labels, SSDToBoundingBoxStepRunner should still work, but just returning null for the bounding box labels
(b) the field should be `static final` not just final (so `public static final String[] COCO_CLASSES_LABELS` on SSDToBoundingBoxStep is better)

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/impl/step/ml/ssd/SSDToBoundingBoxRunner.java
```diff
@@ -52,37 +56,41 @@ public Data exec(Context ctx, Data data) {
...
             double y1 = bArr[0][i][0];
             double x1 = bArr[0][i][1];
             double y2 = bArr[0][i][2];
             double x2 = bArr[0][i][3];
             double p = pArr[0][i];
-            if(p >= threshold) {
-                l.add(BoundingBox.createXY(x1, x2, y1, y2, null, p));
+            float label = lArr[0][0];
+            if (p >= threshold) {
+                l.add(BoundingBox.createXY(x1, x2, y1, y2, COCO_CLASSES_LABELS[(int) label - 1], p));
```
AlexDBlack, line 79: We should add a label index check here, and throw a helpful exception like: "Predicted label index was 30 but only 10 labels were provided"

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/impl/step/bbox/filter/BoundingBoxFilterStepRunner.java
```diff
@@ -0,0 +1,81 @@
...
+import lombok.AllArgsConstructor;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+@AllArgsConstructor
+public class BoundingBoxFilterStepRunner implements PipelineStepRunner {
+
+    protected final BoundingBoxFilterStep step;
```
AlexDBlack, line 35: Let's replace the `@AllArgsConstructor` with manually created contructor (i.e., `public BoundingBoxFilterStepRunner(BoundingBoxFilterStep step){ ...}` and validate that the classesToKeep setting is not null and has length > 0.
Otherwise if a user forgets to set the classes, we could filter everything and they won't know why their pipeline doesn't work.

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/impl/step/bbox/filter/BoundingBoxFilterStepRunner.java
```diff
@@ -0,0 +1,81 @@
...
+        String outName = step.outputName();
+        if (outName == null)
+            outName = BoundingBoxFilterStep.DEFAULT_OUTPUT_NAME;
+
+        Data d = Data.singletonList(outName, boundingBoxes, ValueType.BOUNDING_BOX);
+
+        if (step.keepOtherValues()) {
+            for (String s : data.keys()) {
+                if (!key.equals(s) && !prob.equals(s) &&!labels.equals(s)) {
+                    d.copyFrom(s, data);
```
AlexDBlack, line 71: Also don't copy the value for the input name

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/impl/step/bbox/filter/BoundingBoxFilterStepRunner.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @Override
+    public Data exec(Context ctx, Data data) {
+        String key = "detection_boxes";     //TODO
+        String prob = "detection_scores";
+        String labels = "detection_classes";
+
+        String[] classesToKeep = step.classesToKeep();
+        List<BoundingBox> boundingBoxes = data
+                .getListBoundingBox(step.inputName)
```
AlexDBlack, line 56: One more thing we can do here: if the input name isn't set, we may be able to guess it.
So something like:
```
String inputName = step.inputName();
if(inputName == null){
    String err = "No input name was set in the BoundingBoxFilterStep configuration and input name could not be guessed based on type"
    DataUtils.inferField(data, ValueType.BoundingBox, true, err + " (multiple keys)", err + " (no List<BoundingBox> values)");
}
```

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/impl/step/ml/ssd/SSDToBoundingBoxRunner.java
```diff
@@ -52,37 +57,43 @@ public Data exec(Context ctx, Data data) {
...
             double x2 = bArr[0][i][3];
             double p = pArr[0][i];
-            if(p >= threshold) {
-                l.add(BoundingBox.createXY(x1, x2, y1, y2, null, p));
+            float label = lArr[0][0];
+            if (step.COCO_CLASSES_LABELS.length > 0 && (int) label > step.COCO_CLASSES_LABELS.length){
+                throw new ArrayIndexOutOfBoundsException(String.format("Predicted label index was %s but only %s labels were provided", (int) label, step.COCO_CLASSES_LABELS.length));
+            }
+            if (p >= threshold) {
+                l.add(BoundingBox.createXY(x1, x2, y1, y2, !ArrayUtils.isEmpty(step.COCO_CLASSES_LABELS) ? step.COCO_CLASSES_LABELS[(int) label - 1] : "no label provided", p));
```
AlexDBlack, line 82: All of these references to the COCO dataset have to be removed entirely from here.
The SSD architecture can be used with any datasets; COCO is just one possible dataset. It's like saying "all image 

So again, I'll repeat my earlier comment:
> Two things to change here
(a) the labels should come from the configuration (i.e., something the user sets in SSDToBoundingBoxStep) - and they should be optinal. If the user doesn't provide a list of class labels, SSDToBoundingBoxStepRunner should still work, but just returning null for the bounding box labels
(b) the field should be static final not just final (so public static final String[] COCO_CLASSES_LABELS on SSDToBoundingBoxStep is better)


This means:
1. We introduce a `protected String[] classLabels` in to SSDToBoundingBoxStep (default value: null)
2. We keep the coco labels ONLY as public static final field in SSDToBoundingBoxStep as a convenience to users - i.e., they can do `new SSDToBoundingBoxStep().classLabels(SSDToBoundingBoxStep.COCO_LABELS)`. It's not used as a default or anything.

Also if the labels aren't provided or are empty: we should simply have a null label for that bounding box.

PR #353, opened 2020-06-08
CLI server dependencies resolution (no uber-JARs for CLI launching)

AlexDBlack:
See ADR here for description and motivation.

Also adds the implementation - creation of an otherwise empty JAR with classpath defined in manifest (META-INF/MANIFEST.MF).

### DIFF
--- a/ADRs/0006_CI_Launching_vs_Build.md
+++ b/ADRs/0006_CI_Launching_vs_Build.md
@@ -109,32 +109,28 @@ Use <some_command> to set default profile or pass "-p <profile>" when launching 
 Launching server using default device profile "CUDA_10.2"
 Acquiring dependencies... done
 
-<usual Koanduti Serving launch info>
+<usual Konduit Serving launch info>
 ```
 
-Note that users need only 2 lines here to go from a brand new system to hosting a model server using the optimal hardware.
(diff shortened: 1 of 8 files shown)


ShamsUlAzeem:
Sorry took me time to review it. I was getting familiar with the build tools that were merged prior to this PR in the build system implementation. 

Overall this is a pretty good approach for starting servers with different dependencies and the manifest file will also keep the clutter away from the actual command that was run. For using this approach with the serve command, we can definitely have profiles and then save them in the default konduit-serving root CLI folder. Currently, we're keeping them inside `~/.konduit-serving`. 

We have a `--classpath` or `-cp` argument for the serve command that we can utilize for this. https://github.com/KonduitAI/konduit-serving/blob/a3cf35993de15f6dabff94bc75049faf26b9faf5/konduit-serving-cli/src/main/java/ai/konduit/serving/cli/launcher/command/ServeCommand.java#L81-L85
We won't have to use that directly, instead we can override it based on the manifest files resolved based on the given or default profiles. We'll just have to include `konduit-serving-build` and other related modules into the CLI for this. After that we won't need to download anything else with the CLI and we won't need any other tool when running the java CLI through any other language like we had for our python package.

PR #356, opened 2020-06-10
Gradle build script generation for all needed targets

alexanderst:
Implement build script generation for all targets addressed by current maven structure.


konduit-serving-build/src/main/java/ai/konduit/serving/build/deployments/RpmDeployment.java
```diff
@@ -0,0 +1,130 @@
...
+@NoArgsConstructor
+public class RpmDeployment implements Deployment, OsSpecific {
+    public static final String DEFAULT_EXE_NAME = "konduit-serving-deployment.rpm";
+    public static final String PROP_OUTPUTDIR = "rpm.outputdir";
+    public static final String PROP_RPMNAME = "rpm.name";
+
+    private String outputDir;
+    private String rpmName;
+    private String version;
+    private Os osName;
```
AlexDBlack, line 44: osName field shouldn't be in the configuration - it should be inferred from Target (Like I've added for Arch)

konduit-serving-build/src/main/java/ai/konduit/serving/build/deployments/RpmDeployment.java
```diff
@@ -0,0 +1,130 @@
...
+        String size;
+        if(outFile.exists()){
+            long bytes = outFile.length();
+            double bytesPerMB = 1024 * 1024;
+            double mb = bytes / bytesPerMB;
+            size = String.format("%.2f", mb) + " MB";
+        } else {
+            size = "<EXE not found>";
+        }
+        sb.append("EXE size:       ").append(size);
```
AlexDBlack, line 101: EXE x3

alexanderst:
> rpm, deb, tar, exe and docker are all packing a near-empty JAR not an uber-jar.
> We need them to package the uber-jar for any of them to be useful. I've added a size check so the tests now fail
> 
UberJarDeployment is a standalone type. So should we build uberjar implicitly for RpmDeployment?

konduit-serving-build/src/main/java/ai/konduit/serving/build/deployments/DockerDeployment.java
```diff
@@ -0,0 +1,100 @@
...
+    public DockerDeployment(String outputDir) {
+        this(outputDir, "ks", Deployment.defaultVersion());
+    }
+
+    public DockerDeployment(@JsonProperty("outputDir") String outputDir, @JsonProperty("rpmName") String imageName,
+                             @JsonProperty("version") String version){
+        this.outputDir = outputDir;
+        this.imageName = imageName;
+        this.version = version;
+    }
```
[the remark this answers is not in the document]
alexanderst: Currently base image is taken from resources. We depend on resources for wrapper files anyway. Should I change it to accept some arbitrary path?

https://github.com/KonduitAI/konduit-serving/blob/b0b5e46cb273e5eef0a952ac6febe7cd136c3afc/konduit-serving-build/src/main/java/ai/konduit/serving/build/build/GradleBuild.java#L75

PR #372, opened 2020-06-17
Add async pipeline ADR [WIP]

AlexDBlack:


AlexDBlack:
BackgroundPipeline... or perhaps AsyncPipeline instead?
Trigger instead of LoopTrigger - yes, that sounds good

> For the setCallBackFunction(Function<Data, Data>) how can this be defined into json configuration?

This wouldn't need to be serialized in JSON, it would be set by the executor upon initialization. So only the Trigger fields, excluding the function, needs to be JSON serializable.

ShamsUlAzeem:
Ah that clears my confusion about setting call back through json. Also, `AsyncPipeline` sounds more appropriate.

PR #373, opened 2020-06-17
ClassifierOutputStep

atuzhykov:


ShamsUlAzeem:
@atuzhykov can you also add some reference documentation in the ClassifierOutputStep using `@Schema(description = "...")`. You'll have the dependencies set with you if you merge from master. It will help out in auto generating swagger documentation for our clients in multiple languages. 

Examples are here for how to use the schema: https://github.com/KonduitAI/konduit-serving/blob/master/konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/impl/step/logging/LoggingPipelineStep.java#L34-L35
https://github.com/KonduitAI/konduit-serving/blob/master/konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/impl/step/logging/LoggingPipelineStep.java#L38-L40
So, basically one `@Schema` annotation for the main class and then one for each configurable field.

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/impl/step/ml/classifier/ClassifierOutputRunner.java
```diff
@@ -0,0 +1,201 @@
...
+            inputName = DataUtils.inferField(data, ValueType.NDARRAY, false, errMultipleKeys, errNoKeys);
+        }
+
+
+        NDArray classifierOutput = data.getNDArray(inputName);
+
+        boolean batch = false;
+        if (classifierOutput.shape().length == 2 && classifierOutput.shape()[0] > 1) {
+            batch = true;
+        }
```
AlexDBlack, line 73: Let's check that rank (shape().length) is 1 or 2. For 3, we should throw an exception like:
```
"Invalid input to ClassifierOutputStep: only rank 1 or 2 inputs are available, got array with shape " + Arrays.toString(classifierOutput().shape())
```

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/impl/step/ml/classifier/ClassifierOutputRunner.java
```diff
@@ -0,0 +1,201 @@
...
+        if (!batch) {
+            double[] classifierOutputArr = squeze(classifierOutput);
+            double[] maxValueWithIdx = getMaxValueAndIndex(classifierOutputArr);
+            double prob = maxValueWithIdx[0];
+            long index = (long) maxValueWithIdx[1];
+            String label = labels.get((int) index);
+
+            if (step.topN() != null && step.topN() > 1) {
+                if (step.returnProb()) {
+                    data.putListDouble(step.probName(), Collections.singletonList(prob));
```
AlexDBlack, line 97: Technically these prob/index/labelName field can be null, in sp
Most of the time they won't, but they could be under the following circumstances:
1.  `new ClassifierOutputStep().probName(null);`
2. Manually written JSON
3. JSON created by one of the other clients (python, javascript, etc)

so: Let's switch to something like this for all of the names:
```
String probName = step.probName() == null ? ClassifierOutputStep.DEFAULT_PROB_NAME : step.probName()
```

And let's do that once earlier (to avoid code duplication). Or a simple method that's used in each place would be fine too.

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/impl/step/ml/classifier/ClassifierOutputRunner.java
```diff
@@ -0,0 +1,201 @@
...
+
+        if (batch) {
+
+            int bS = (int) classifierOutput.shape()[1];
+            double[][] y = classifierOutput.getAs(double[][].class);
+
+            List<Double> probs = new ArrayList<Double>();
+            List<Long> indeces = new ArrayList<Long>();
+            List<String> labelsList = new ArrayList<String>();
+            List<NDArray> allPropabilities = new ArrayList<NDArray>();
```
AlexDBlack, line 131: We'll use `List<List<Double>>` for consistency with List<Double> for the non-batch case
so `double[][] arr = ...` then loop over each `arr[i]` (each row)

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/impl/step/ml/classifier/ClassifierOutputRunner.java
```diff
@@ -0,0 +1,201 @@
...
+        if (arr.shape().length == 1) {
+            return arr.getAs(double[].class);
+        }
+
+        // i.e we have [1, numClasses] array
+        if (arr.shape().length == 2 && arr.shape()[0] == 1) {
+            return arr.getAs(double[][].class)[0];
+        }
+
+        return null;
```
AlexDBlack, line 179: I don't think this will ever occur, but it should probably be an exception?

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/impl/step/ml/classifier/ClassifierOutputStep.java
```diff
@@ -0,0 +1,100 @@
...
+
+
+import java.util.List;
+
+
+@Builder
+@Data
+@Accessors(fluent = true)
+@AllArgsConstructor
+@JsonName("CLASSIFIER")
```
AlexDBlack, line 37: Let's maybe call this `CLASSIFIER_OUTPUT`
Just `CLASSIFIER` might make people think it's the thing that does the classification (i.e., people might think it's a ML model)

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/impl/step/ml/classifier/ClassifierOutputStep.java
```diff
@@ -0,0 +1,100 @@
...
+        //Normally this would be unnecessary to set default values here - but @Builder.Default values are NOT treated as normal default values.
+        //Without setting defaults here again like this, the fields would actually be null
+        this.returnLabel = true;
+        this.returnIndex = true;
+        this.returnProb = true;
+        this.indexName = DEFAULT_OUTPUT_INDEX_NAME;
+        this.labelName = DEFAULT_OUTPUT_LABEL_NAME;
+        this.probName = DEFAULT_OUTPUT_PROB_NAME;
+        this.topN = null;
+        this.allProbabilities = false;
```
AlexDBlack, line 96: I've been a bit inconsistent on this (I need to do a pass to make this consistent throughout KS), but let's remove the `@Builder` and `@Builder.Default` (and the no-arg constructor) and rely on `@Accessors(fluent = true)` only.

In practice for users that means this:
```
ClassifierOutputStep.builder()
    .someMethod(...)
    .build()
```
becomes this:
```
new ClassifierOutputStep()
    .someMethod(...)
```

We avoid the lombok builder defaults issue with that too

konduit-serving-pipeline/src/test/java/ai/konduit/serving/pipeline/impl/step/TestClassifierOutputStep.java
```diff
@@ -0,0 +1,85 @@
+package ai.konduit.serving.pipeline.impl.step;
```
AlexDBlack, line 1: Copyright

konduit-serving-pipeline/src/test/java/ai/konduit/serving/pipeline/impl/step/TestClassifierOutputStep.java
```diff
@@ -0,0 +1,85 @@
...
+                                                        .topN(topN)
+                                                        .indexName("index")
+                                                        .probName("prob")
+                                                        .Labels(labels))
+                                                .build();
+
+
+                                        NDArray preds = NDArray.create(values);
+                                        Data in = Data.singleton("preds", preds);
+                                        Data out = p.executor().exec(in);
```
AlexDBlack, line 70: Let's add some checks like:
```
if(retProb)
    assertTrue(out.has("prob") && out.type("prob") == ValueType.DOUBLE);
```

Let's also add a test for float[] values too - I suspect the NDArray.getAs(double[][].class) in your step might not like that... but that's something we should fix generally, not just for this PR//step.

But otherwise :+1:

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/impl/step/ml/classifier/ClassifierOutputStep.java
```diff
@@ -0,0 +1,93 @@
...
+    public static final String DEFAULT_LABEL_NAME = "label";
+
+
+    @Schema(description = "inputName - optional. If set: this represents the NDArray. If not set: use DataUtils.inferField to find an NDArray field")
+    String inputName;
+
+
+
+    @Schema(description = "returnLabel, default is true; if false, don't return label")
+    boolean returnLabel = true;
```
ShamsUlAzeem, line 63: The annotation can also define a default value here with `defaultValue`. For a boolean we can specify a string instead. like `defaultValue = "true"`

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/impl/step/ml/classifier/ClassifierOutputStep.java
```diff
@@ -0,0 +1,93 @@
...
+        "- The predicted class index - as an integer (long)<br>" +
+        "- The predicted class probability - as a Double")
+public class ClassifierOutputStep implements PipelineStep {
+    public static final String DEFAULT_PROB_NAME = "prob";
+    public static final String DEFAULT_INDEX_NAME = "index";
+    public static final String DEFAULT_LABEL_NAME = "label";
+
+
+    @Schema(description = "inputName - optional. If set: this represents the NDArray. If not set: use DataUtils.inferField to find an NDArray field")
+    String inputName;
```
ShamsUlAzeem, line 58: Might want to add `private` access modifiers with all of the configurable variables we're defining here cuz I'm not sure if the serializer is gonna pick up package private variables

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/impl/step/ml/classifier/ClassifierOutputStep.java
```diff
@@ -0,0 +1,93 @@
...
+    @Schema(description = "returnIndex, default is true")
+    boolean returnIndex = true;
+
+
+    @Schema(description = " returnProb, default is true")
+    boolean returnProb = true;
+
+
+    @Schema(description = "output names for the labels")
+    String labelName = DEFAULT_LABEL_NAME;
```
ShamsUlAzeem, line 75: Default value here as well: `defaultValue = "prob"` or `defaultValue = DEFAULT_LABEL_NAME`

PR #379, opened 2020-06-18
Support of remote configs

alexanderst:


konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/api/protocol/URIResolver.java
```diff
@@ -0,0 +1,49 @@
...
+            return true;
+        }
+        return false;
+    }
+
+    private static File cacheDirectory;
+
+    public static File getCacheDirectory() {
+        if (cacheDirectory == null) {
+            File f = new File(System.getProperty("user.home"), ".konduit_cache/");
```
AlexDBlack, line 27: Let's also make this configurable (this is a good default though). Doing that in the next PR is fine.

konduit-serving-pipeline/src/test/java/ai/konduitai/serving/pipeline/protocol/TestURL.java
```diff
@@ -0,0 +1,121 @@
...
+
+        url = HTTPS + HOST + ":" + PORT + "/src/test/resources/config/config.json";
+        data = FileUtils.readFileToString(URIResolver.getFile(url), "UTF-8");
+        assertTrue(data.contains("pipelineSteps"));
+    }
+
+    @Ignore("Windows bound manual test")
+    @Test
+    public void testFTP() throws IOException, InterruptedException {
+        FakeFtpServer fakeFtpServer = new FakeFtpServer();
```
AlexDBlack, line 68: If this is windows-specific (that's not ideal though) we can use `assumeTrue(SystemUtils.IS_OS_WINDOWS)` (that's in commons lang 3) - it'll dynamically skip non-Windows.

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/api/protocol/URIResolver.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public static File getFile(String uri) throws IOException {
+
+        String fileName = FilenameUtils.getName(uri);
+        File cachedFile = new File(getCacheDirectory(), fileName);
+        if (cachedFile.exists()) {
+            return cachedFile;
+        }
+
+        URI u = URI.create(uri);
```
AlexDBlack, line 43: As a fallback - let's assume anything without a URI scheme (i.e., no `xyz://`) is a file... that allows users to do things like `C:/my/pipeline.json` etc instead of needing to do `file:///my/pipeline.json` all the time.

konduit-serving-build/src/main/java/ai/konduit/serving/build/cli/BuildCLI.java
```diff
@@ -180,6 +181,11 @@ public void exec(String[] args) throws Exception {
         if(config != null){
             Map<String,String> props = new HashMap<>();
             for(String s : config){
+                if (RemoteUtils.isUrl(s))
+                    continue;
+                if (s.startsWith("http://")) {
+                    s = RemoteUtils.configFromHttp(s);
+                }
```
AlexDBlack, line 188: Not sure on the `if(RemoteUtils.isUrl(s))` part. We're skipping URLs but then downloading them on the next section?

Also we should probably use URIResolver.getFile here instead.

PR #402, opened 2020-06-23
New CLI extension: Serve Command with Manifest JAR file

ShamsUlAzeem:
Implementing ADR: https://github.com/KonduitAI/konduit-serving/blob/master/ADRs/0006-CI_Launching_vs_Build.md


konduit-serving-cli/src/main/java/ai/konduit/serving/cli/launcher/command/build/extension/ProfileCommand.java
```diff
@@ -0,0 +1,509 @@
...
+        } else {
+            saveProfile(profileName, fillProfileValues(new Profile()));
+        }
+    }
+
+    private void editProfile() {
+        if(isProfileExists(profileName)) {
+            if(profileName.equals("CPU") || profileName.equals("CUDA")) {
+                out.format("Cannot edit pre-set profiles with name 'CPU' or 'CUDA'.%n");
+                System.exit(1);
```
AlexDBlack, line 193: We should move these contants ("CPU", "CUDA", "x86" etc) to `static final String` fields on the class.

Also, we should probably allow editing of these CPU and CUDA profiles, for two reasons:
(a) If the user can't run AVX2 for example and wants to default to x86 only, or
(b) If the user installs a different CUDA version after KS installation

konduit-serving-cli/src/main/java/ai/konduit/serving/cli/launcher/command/build/extension/ProfileCommand.java
```diff
@@ -0,0 +1,509 @@
...
+
+        if(operatingSystem != null) {
+            profile.operatingSystem(operatingSystem);
+        }
+        if(computeDevice != null) {
+            profile.computeDevice(computeDevice);
+        }
+
+        if(StringUtils.containsIgnoreCase(profile.computeDevice(), "cuda")) {
+            profile.cpuArchitecture("x86");
```
AlexDBlack, line 215: arm64 and ppc64le both have CUDA support so we shouldn't hardcode x86 here

konduit-serving-cli/src/main/java/ai/konduit/serving/cli/launcher/command/build/extension/ProfileCommand.java
```diff
@@ -0,0 +1,509 @@
...
+
+        if(serverTypes != null && !serverTypes.isEmpty()) {
+            profile.serverTypes(serverTypes);
+        }
+        if(additionalDependencies != null && !additionalDependencies.isEmpty()) {
+            profile.additionalDependencies(additionalDependencies);
+        }
+
+        if(StringUtils.containsIgnoreCase(profile.computeDevice(), "cuda")) {
+            String cudaVersion = profile.computeDevice().split("_")[1].trim();
```
AlexDBlack, line 226: We should validate format. If user passes in "cuda10.0" we'll get an exception on this line

konduit-serving-cli/src/main/java/ai/konduit/serving/cli/launcher/command/build/extension/ProfileCommand.java
```diff
@@ -0,0 +1,509 @@
...
+            String cudaRedistPackage = null;
+            Pair<String, String> cudaInstall = findCudaInstall();
+            if(cudaInstall == null) {
+                switch (cudaVersion) {
+                    case "10.0":
+                        out.format("No CUDA install found and no available redist package for cuda version: '%s' found. " +
+                                "Make sure to install CUDA from: %s before starting a konduit server.%n", cudaVersion,
+                                "https://developer.nvidia.com/cuda-10.0-download-archive");
```
AlexDBlack, line 234: We should probably explicitly mention the user passed CUDA 10.0 and that they could use CUDA 10.1 or 10.2 as an alternative.

konduit-serving-cli/src/main/java/ai/konduit/serving/cli/launcher/command/build/extension/ProfileCommand.java
```diff
@@ -0,0 +1,509 @@
...
+        List<GraphicsCard> nvidiaGraphicsCard = new ArrayList<>();
+        for(GraphicsCard graphicsCard : systemInfo.getHardware().getGraphicsCards()) {
+            String vendor = graphicsCard.getVendor();
+            if (vendor != null && StringUtils.containsIgnoreCase(vendor, "nvidia")) {
+                nvidiaGraphicsCard.add(graphicsCard);
+            }
+        }
+
+        if(!nvidiaGraphicsCard.isEmpty() && !cpuProfile.operatingSystem().equalsIgnoreCase("mac")) {
+            log.error("Found the following cuda compatible devices in the local system: {}", nvidiaGraphicsCard);
```
AlexDBlack, line 324: Should be log.info?

konduit-serving-cli/src/main/java/ai/konduit/serving/cli/launcher/command/build/extension/ProfileCommand.java
```diff
@@ -0,0 +1,509 @@
...
+
+            String defaultCudaVersion =  "CUDA_10.1";
+            Profile cudaProfile = new Profile(defaultCudaVersion,
+                    "x86", cpuProfile.operatingSystem(), Arrays.asList("HTTP", "GRPC"),
+                    Collections.singletonList("ch.qos.logback:logback-classic:1.2.3"));
```
AlexDBlack, line 329: Again we can't/shouldn't assume x86 here. And let's not have these default values here.
Ideally they should be centralized - something like `Profile.withDefaults(name, arch, os)`

konduit-serving-cli/src/main/java/ai/konduit/serving/cli/launcher/command/build/extension/ProfileCommand.java
```diff
@@ -0,0 +1,509 @@
...
+                        return parentFile.getAbsolutePath();
+                    }
+                    break;
+                }
+                line = bufferedReader.readLine();
+            }
+        }
+
+        return null;
+    }
```
AlexDBlack, line 508: Another thing we can do is check the default CUDA install locations too. But if it's not on the path it may not be loadable anyway... So maybe we look at that in a separate PR (it'l take a bit of testing to ensure it works reliably)

konduit-serving-cli/src/main/java/ai/konduit/serving/cli/launcher/command/build/extension/ServeBuildCommand.java
```diff
@@ -0,0 +1,174 @@
...
+                Object buildCli = buildCliClass.getConstructor().newInstance();
+                Method buildCliMainMethod = buildCliClass.getMethod("exec", String[].class);
+                buildCliMainMethod.setAccessible(true);
+
+                Profile profile = profileName != null ? ProfileCommand.getProfile(profileName) : ProfileCommand.getDefaultProfile();
+                if(profile == null) {
+                    if(profileName == null) {
+                        out.println("Couldn't find a default profile. Starting server with 'CPU' profile");
+                    } else {
+                        out.format("Couldn't find a profile with the specified name: '%s'. Starting server with 'CPU' profile.%n", profileName);
```
AlexDBlack, line 90: We should probably just fail on this case instead (to avoid user being surprised by the behaviour after a profile name typo, for example)

konduit-serving-cli/src/main/java/ai/konduit/serving/cli/launcher/command/build/extension/ServeBuildCommand.java
```diff
@@ -0,0 +1,174 @@
...
+                    this.classpath = mfJar.getAbsolutePath();
+                } else {
+                    this.classpath += File.pathSeparator + mfJar.getAbsolutePath();
+                }
+            }
+        } catch (IOException e) {
+            log.error("Unable to write build pipeline data to {}.", savePath.getAbsolutePath(), e);
+            System.exit(1);
+        } catch (ClassNotFoundException e) {
+            out.println("Unable to find classes for building manifest jar. Continuing without runtime build...");
```
AlexDBlack, line 128: If the manifest JAR build fails, we likely can't launch the server successfully (due to missing modules/dependencies), so probably should just exit.

konduit-serving-data/konduit-serving-nd4j/src/main/java/ai/konduit/serving/data/nd4j/Nd4jModuleInfo.java
```diff
@@ -31,11 +31,11 @@
...
                 @Dependency(gId = "org.nd4j", aId = "nd4j-cuda-10.0", ver = "1.0.0-beta7", classifier = {"linux-x86_64", "linux-ppc64le", "linux-arm64","windows-x86_64"})}),
-        @Requires(requires = Req.ALL, value = {@Dependency(gId = "org.nd4j", aId = "nd4j-cuda-10.1", ver = "1.0.0-beta7"),
+        @Requires(requires = Req.ALL, value = {@Dependency(gId = "org.nd4j", aId = "nd4j-cuda-10.1-platform", ver = "1.0.0-beta7"),
                 @Dependency(gId = "org.nd4j", aId = "nd4j-cuda-10.1", ver = "1.0.0-beta7", classifier = {"linux-x86_64", "linux-ppc64le", /*"linux-arm64",*/ "windows-x86_64"})}),      //Note 1.0.0-beta7 was only released for linux-arm64 for CUDA 10.0
-        @Requires(requires = Req.ALL, value = {@Dependency(gId = "org.nd4j", aId = "nd4j-cuda-10.2", ver = "1.0.0-beta7"),
+        @Requires(requires = Req.ALL, value = {@Dependency(gId = "org.nd4j", aId = "nd4j-cuda-10.2-platform", ver = "1.0.0-beta7"),
                 @Dependency(gId = "org.nd4j", aId = "nd4j-cuda-10.2", ver = "1.0.0-beta7", classifier = {"linux-x86_64", "linux-ppc64le", /*"linux-arm64",*/ "windows-x86_64"})})
```
AlexDBlack, line 39: Again, no -platform

konduit-serving-vertx/src/main/java/ai/konduit/serving/vertx/settings/DirectoryFetcher.java
```diff
@@ -48,6 +48,34 @@ public static File getWorkingDir() {
...
+    /**
+     * Creates the profiles directory if missing and fetches it.
+     * @return konduit-serving profiles directory.
+     */
+    public static File getProfilesDir() {
+        return createAndValidateDirectory(
+                fetchValueBasedOnPriority(
+                        System.getenv(EnvironmentConstants.PROFILES_DIR),
+                        System.getProperty(PropertiesConstants.PROFILES_DIR),
+                        getDefaultProfilesDir()
```
AlexDBlack, line 74: For build dir - we should probably put it in a subdirectory (not just unique filenames as you've done in ServeBuildCommand) to avoid the risk of gradle's files causing problems due to the same location for different launches (or between concurrent launches).

AlexDBlack:
Looks like the upgraded OSHI version is a problem for the old API... I'm seeing this during tests on command line.
https://gist.github.com/AlexDBlack/3354f3c424b59db7f99bb364229ca148

AlexDBlack:
**Changes I've made here:**
1. Make CLI an uber-jar
2. Add logback dependency to CLI uberjar
3. Add oshi-core to CLI uberjar
4. Profile: @Data annotation
5. Fixed an infinite loop when user calls `konduit profile list` as the first command - https://gist.github.com/AlexDBlack/983d84862ead34ade15829d86677cb59
6. Small tweaks to messages
7. Make konduit-serving-build module a direct dependency of konduit-serving-cli (no reflection, and so we only have to distribute one JAR)



**Some identified problems (not yet fixed)**
Error with empty profile directory or no profile directory + "konduit serve ..."
https://gist.github.com/AlexDBlack/a4d159787a487d1f26303924a378c465

"konduit profile list" (first run) https://gist.github.com/AlexDBlack/a8b93854391483b9be5c32152f6a0bcb
- VRAM is wrong: `vRam=4293918720` - it's a 12GB card (reporting just under 4GB)
- "No profile with name CUDA exists." (but we're in the process of creating it here?)
- Profiles are printed twice

If inference configuration doesn't exist (bad path) - NPE - https://gist.github.com/AlexDBlack/ecba7407fe00305654d8d669fe760de6

If manifest JAR build fails, we get some bad exceptions - https://gist.github.com/AlexDBlack/66ba86245d18bfcc5367df94a6a36729
(the class not found exception part is irrelevant now but other failures could still cause this problem)

"Unable to get default profile" + manifest JAR present via classpath arg but clearly not working
https://gist.github.com/AlexDBlack/db84e256c7f697522a633a3b4ed36fb1
(here's code used to create test file - https://gist.github.com/AlexDBlack/3e6b55f7c60daf225b414e3656fc434a )

konduit-serving-cli/src/main/java/ai/konduit/serving/cli/launcher/command/build/extension/ProfileCommand.java
ShamsUlAzeem: They are validated when setting through the profile command arguments using the json setter: https://github.com/KonduitAI/konduit-serving/blob/6b2bdca23ebb45bb0ba6f6888864ce6900fc413f/konduit-serving-cli/src/main/java/ai/konduit/serving/cli/launcher/command/build/extension/model/Profile.java#L99-L105

Also, when setting them through the constructors: https://github.com/KonduitAI/konduit-serving/blob/6b2bdca23ebb45bb0ba6f6888864ce6900fc413f/konduit-serving-cli/src/main/java/ai/konduit/serving/cli/launcher/command/build/extension/model/Profile.java#L91

konduit-serving-cli/src/main/java/ai/konduit/serving/cli/launcher/command/build/extension/ProfileCommand.java
ShamsUlAzeem: Issue created here: https://github.com/KonduitAI/konduit-serving/issues/436

ShamsUlAzeem:
Fixed most issues and created new issues for some of them. Oshi hardware gives wrong information. Might be some setting that I don't know about. I've created issue for it here: https://github.com/KonduitAI/konduit-serving/issues/438

konduit-serving-build/pom.xml
```diff
@@ -210,7 +210,7 @@
                         </configuration>
                     </execution>
                 </executions>
-            </plugin>
+            </plugin>-->
```
AlexDBlack, line 213: We should remove the maven shade plugin entirely from this pom. If users need to call the build tool CLI they can do it from the konduit-serving-cli uberjar instead.
(And yes, this commenting-out was my change, I probably should have removed instead of commenting :))

konduit-serving-cli/src/main/java/ai/konduit/serving/cli/launcher/command/ListCommand.java
```diff
@@ -120,8 +121,9 @@ private void printServerDetails(int index, String printFormat, String id, String
 
         try {
             configuration = FileUtils.readFileToString(new File(DirectoryFetcher.getServersDataDir(), pid + ".data"), StandardCharsets.UTF_8);
-            InferenceConfiguration inferenceConfiguration = InferenceConfiguration.fromJson(configuration);
-            hostAndPort = String.format("%s:%s", inferenceConfiguration.getHost(), inferenceConfiguration.getPort());
+            JsonNode jsonNode = ObjectMappers.json().readTree(configuration);
+
+            hostAndPort = String.format("%s:%s", jsonNode.get("host"), jsonNode.get("port")).replaceAll("\"", "");
```
AlexDBlack, line 126: What was the reasoning for getting it from the JSON instead of the inference configuration?
(I mean I think both should work, just wondering why the change)

konduit-serving-cli/src/main/java/ai/konduit/serving/cli/launcher/command/build/extension/ProfileCommand.java
```diff
@@ -0,0 +1,554 @@
...
+        List<GraphicsCard> nvidiaGraphicsCard = new ArrayList<>();
+        for(GraphicsCard graphicsCard : new SystemInfo().getHardware().getGraphicsCards()) {
+            String vendor = graphicsCard.getVendor();
+            if (vendor != null && StringUtils.containsIgnoreCase(vendor, "nvidia")) {
+                nvidiaGraphicsCard.add(graphicsCard);
+            }
+        }
+
+        if(!nvidiaGraphicsCard.isEmpty() && !cpuProfile.operatingSystem().equalsIgnoreCase(Profile.OperatingSystem.MAC.name())) {
+            log.info("Found the following cuda compatible devices in the local system: {}", nvidiaGraphicsCard);
```
AlexDBlack, line 330: Tiny nitpick: `cuda` -> `CUDA`

konduit-serving-cli/src/main/java/ai/konduit/serving/cli/launcher/command/build/extension/ProfileCommand.java
```diff
@@ -0,0 +1,554 @@
...
+
+                addDependencies(cudaProfile, CUDA_10_1_REDIST_DEPENDENCY);
+            }
+
+            profiles.put(DEFAULT_CUDA_PROFILE_NAME, cudaProfile);
+
+            saveProfiles(profiles);
+            setDefaultProfile(DEFAULT_CUDA_PROFILE_NAME);
+        } else {
+            log.info("No cuda compatible devices found in the current system.");
```
AlexDBlack, line 354: `cuda` -> `CUDA`

konduit-serving-cli/src/main/java/ai/konduit/serving/cli/launcher/command/build/extension/ProfileCommand.java
```diff
@@ -0,0 +1,554 @@
...
+                }
+                line = bufferedReader.readLine();
+            }
+        }
+
+        return null;
+    }
+
+    private static String findCudaInstallPath(String mainCommandName, String cudaVersion) throws IOException {
+        try (BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(new ProcessBuilder(Arrays.asList("where", mainCommandName)).start().getInputStream()))) {
```
AlexDBlack, line 538: Shouldn't this be "where" on windows, "which" on linux?

konduit-serving-cli/src/main/java/ai/konduit/serving/cli/launcher/command/build/extension/model/Profile.java
```diff
@@ -0,0 +1,185 @@
...
+import lombok.Data;
+import org.apache.commons.lang3.SystemUtils;
+import org.nd4j.shade.jackson.annotation.JsonGetter;
+import org.nd4j.shade.jackson.annotation.JsonSetter;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+@Data
```
AlexDBlack, line 31: `@Data` generates getters/setters, which we probably don't want - i.e., no point having a `getComputeDevice()` and a `computeDevice()`.
So probably just `@EqualsAndHashCode` and `@ToString` needed here

AlexDBlack:
Feedback after testing

In help (-h), <subCommand> should be first in list, <profile_name> should be second
https://gist.github.com/AlexDBlack/c2797c710dd3d75f2215fb490eb0a981

unknown subcommands give ugly exception
https://gist.github.com/AlexDBlack/2e9dbef7bfdec1b7d0da4e3c8b035e39
(this should have been --help or -h)

unknown server type gives ugly exception
https://gist.github.com/AlexDBlack/8a16e8cf183482c7683ab3e06b276170

edit doesn't seem to work, at least for this case I tested:
https://gist.github.com/AlexDBlack/5ca9655c7ff86279f1a9638f9755ba83

not exactly a new thing here - but we should handle missing PID files more gracefully: https://gist.github.com/AlexDBlack/b74ea52ae491ef00a61f08b78a5b7e05

Now that the build tool is no longer an uberjar, we probably need a top-level 'build' command that calls the build CLI, for creating uberjars, docker images, etc
i.e., `konduit build -p myconfig.json -o WINDOWS` etc just passes the args to ai.konduit.serving.build.cli.BuildCLI


As for the CUDA NoAvailableBackendException issue - the backend (and classifier JAR) is definitely included in the manifest, so I'm not sure why we're getting that :(
I'll keep looking into it.

Issue #390, opened 2020-06-22
API consistency - standardize on fluent API (remove builder and non-fluent getters)


AlexDBlack:
For the most part, I have been writing pipeline steps with a fluent API with no builder, i.e., with the following Lombok annotations:
```
@Data
@Accessors(fluent=true)
```
For example: https://github.com/KonduitAI/konduit-serving/blob/master/konduit-serving-data/konduit-serving-image/src/main/java/ai/konduit/serving/data/image/step/crop/ImageCropStep.java#L46-L47

However, some have `@Builder` also, such as:
https://github.com/KonduitAI/konduit-serving/blob/master/konduit-serving-data/konduit-serving-image/src/main/java/ai/konduit/serving/data/image/step/bb/draw/DrawBoundingBoxStep.java#L42

---
---

What we should do:
1. Ensure we have `@Data @Accessors(fluent=true)` on all pipeline steps and configuration type classes (such as ImageToNDArrayConfig)

2. Remove `@Builder` from PipelineStep implementations and other config classes. In doing so, we will need 2 additional changes:
- Remove all of the "`@Builder` workaround" no-arg constructors like this one: https://github.com/KonduitAI/konduit-serving/blob/master/konduit-serving-data/konduit-serving-image/src/main/java/ai/konduit/serving/data/image/convert/ImageToNDArrayConfig.java#L136-L145
- Update any tests using builders to fluent (i.e., `MyClass.builder().someConfig(...).build()` --> `new MyClass().someConfig(...)`

PR #404, opened 2020-06-24
Add DrawPointsStep and DrawHeatmapStep

treo:
DrawPointsStep takes a point or a list of points and a mapping for their labels to colors, and draws those on a black background. 

DrawHeatmapStep takes a point or a list of points and draws a heatmap. It collects point positions over time, so depending on the fading factor, old "heat" is retained.

Both steps pass through all data that is given to them - I'm not sure if that is something that they should do, as not all steps actually do that. On some it is a configuration option, others just don't do it.

And another question on style: As we have lombok in the project, we can use `val`, so far I've tried to stay as close to plain java as possible. What is the preferred style there?

### DIFF
--- a/konduit-serving-data/konduit-serving-image/src/main/java/ai/konduit/serving/data/image/step/point/draw/DrawPointsStep.java
+++ b/konduit-serving-data/konduit-serving-image/src/main/java/ai/konduit/serving/data/image/step/point/draw/DrawPointsStep.java
@@ -1,0 +1,86 @@
+/*
+ *  ******************************************************************************
+ *  * Copyright (c) 2020 Konduit K.K.
+ *  *
+ *  * This program and the accompanying materials are made available under the
+ *  * terms of the Apache License, Version 2.0 which is available at
+ *  * https://www.apache.org/licenses/LICENSE-2.0.
+ *  *
+ *  * Unless required by applicable law or agreed to in writing, software
+ *  * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ *  * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ *  * License for the specific language governing permissions and limitations
+ *  * under the License.
+ *  *
+ *  * SPDX-License-Identifier: Apache-2.0
+ *  *****************************************************************************
+ */
+
+package ai.konduit.serving.data.image.step.point.draw;
+
+import ai.konduit.serving.annotation.json.JsonName;
+import ai.konduit.serving.pipeline.api.step.PipelineStep;
+import io.swagger.v3.oas.annotations.media.Schema;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.Singular;
+import lombok.experimental.Accessors;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Draw 2D points<br>
+ * Configuration:<br>
+ * <ul>
+ *     <li><b>classColors</b>: Optional: A list of colors to use for each class. Must be in one of the following formats: hex/HTML - "#788E87",
+ *             RGB - "rgb(128,0,255)", or one of 16 HTML color name such as \"green\" (https://en.wikipedia.org/wiki/Web_colors#HTML_color_names).
+ *             If no colors are specified, or not enough colors are specified, random colors are used instead (note consistent between runs).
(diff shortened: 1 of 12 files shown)


konduit-serving-data/konduit-serving-image/src/main/java/ai/konduit/serving/data/image/step/point/draw/DrawPointsStepRunner.java
```diff
@@ -0,0 +1,152 @@
...
+            width = step.width();
+            height = step.height();
+        }else{
+            throw new IllegalArgumentException("You have to provide either a reference image or width AND height!");
+        }
+
+        // turn points with relative addressing to absolute addressing
+        List<Point> absPoints = new ArrayList<>(points.size());
+        for (Point point : points) {
+            absPoints.add(point.toAbsolute(width, height));
```
AlexDBlack, line 115: We should probably mention that it accepts both types of points in the doc

konduit-serving-data/konduit-serving-image/src/main/java/ai/konduit/serving/data/image/step/point/draw/DrawPointsStepRunner.java
```diff
@@ -0,0 +1,152 @@
...
+        }
+
+        // create empty image
+        Mat image = new Mat();
+        image.put(Mat.zeros(height, width, CvType.CV_8UC3));
```
AlexDBlack, line 120: As mentioned earlier - let's allow drawing on the image as an option too

konduit-serving-data/konduit-serving-image/src/main/java/ai/konduit/serving/data/image/step/point/heatmap/DrawHeatmapStep.java
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * @author Paul Dubs
+ */
+@Builder
+@Data
+@Accessors(fluent = true)
+@AllArgsConstructor
+@JsonName("DRAW_HEATMAP")
+@Schema(description = "A pipeline step that configures how to draw 2D points, optionally on an image.")
+public class DrawHeatmapStep implements PipelineStep {
```
AlexDBlack, line 54: Again, "on image" vs. "image as size reference".
Being able to overlay on an image would definitely be a nice addition (at the very least, we'll open an issue and add it later)

konduit-serving-data/konduit-serving-image/src/main/java/ai/konduit/serving/data/image/step/point/heatmap/DrawHeatmapStep.java
```diff
@@ -0,0 +1,79 @@
...
+    public static final String DEFAULT_OUTPUT_NAME = "image";
+
+    @Schema(description = "Names of the points to be used for the heatmap. Accepts both single points and lists of points.")
+    @Singular
+    private List<String> points;
```
AlexDBlack, line 59: Same docs clarification issue

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/impl/step/bbox/point/BoundingBoxToPointStep.java
```diff
@@ -35,6 +35,7 @@
  * the output will be a single value; if the input is a list, the output will be a list.<br>
  */
 @Data
+@Builder
```
AlexDBlack, line 38: No builder: we're standardizing on fluent API everywhere - https://github.com/KonduitAI/konduit-serving/issues/390

konduit-serving-data/konduit-serving-image/src/main/java/ai/konduit/serving/data/image/step/point/draw/DrawPointsStep.java
```diff
@@ -0,0 +1,87 @@
...
+ * </ul>
+ *
+ * @author Paul Dubs
+ */
+@Builder
+@Data
+@Accessors(fluent = true)
+@AllArgsConstructor
+@JsonName("DRAW_POINTS")
+@Schema(description = "A pipeline step that configures how to draw 2D points, optionally on an image.")
```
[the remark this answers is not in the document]
treo: Ah, forgot to change it here.

I initially wanted to draw on top of the image, but decided against it, the main reason being that image composition should probably be a step of its own.

PR #431, opened 2020-07-01
json/yaml coverage tests 

atuzhykov:


konduit-serving-cli/src/test/java/ai/konduit/serving/cli/data/JsonCoverageTest.java
```diff
@@ -0,0 +1,28 @@
+package ai.konduit.serving.cli.data;
```
AlexDBlack, line 1: Copyright header (here and all the other new classes)

konduit-serving-cli/src/test/java/ai/konduit/serving/cli/data/JsonCoverageTest.java
```diff
@@ -0,0 +1,28 @@
...
+        return ObjectMappers.fromJson(json, c);
+    }
+
+    @Override
+    public Object fromYaml(Class<?> c, String yaml) {
+        return ObjectMappers.fromYaml(yaml, c);
+    }
+
+
+
```
AlexDBlack, line 25: For all modules (currently) without anything that extends TextConfig (i.e., any JsonCoverageTest without an `@Test`) I think we'll need at least one test to be implemented in the class for the `@AfterClass` to run.
Something like this should suffice:
```
@Test
public void test(){
    //Empty test to ensure @Before is run at least one and @AfterClass is run
}
```

konduit-serving-common-tests/pom.xml
```diff
@@ -58,5 +58,38 @@
             <artifactId>jetty-server</artifactId>
             <version>9.4.19.v20190610</version>
         </dependency>
+        <dependency>
+            <groupId>org.reflections</groupId>
+            <artifactId>reflections</artifactId>
+            <version>0.9.9</version>
```
AlexDBlack, line 64: Don't use hardcoded versions like this. Versions should go in the parent pom - there's already a version defined.
Use `<version>${reflections.version}</version>` instead

konduit-serving-common-tests/pom.xml
```diff
@@ -58,5 +58,38 @@
...
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.codehaus.groovy</groupId>
+            <artifactId>groovy-all</artifactId>
+            <version>2.4.8</version>
+            <scope>compile</scope>
+        </dependency>
```
AlexDBlack, line 72: This groovy dependency should not be necessary.
Also vertx-core should not be necessary.

Also remove `<scope>compile</scope>` from these dependencies that remain (compile scope is the default so specifying it as such is unnecessary)

konduit-serving-data/konduit-serving-image/pom.xml
```diff
@@ -76,6 +76,12 @@
             <version>${nd4j.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>ai.konduit.serving</groupId>
+            <artifactId>konduit-serving-common-tests</artifactId>
+            <version>0.1.0-SNAPSHOT</version>
```
AlexDBlack, line 82: Use `<version>${project.version}</version>`

konduit-serving-data/konduit-serving-nd4j/pom.xml
```diff
@@ -68,6 +68,12 @@
             <version>${logback.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>ai.konduit.serving</groupId>
+            <artifactId>konduit-serving-common-tests</artifactId>
+            <version>0.1.0-SNAPSHOT</version>
```
AlexDBlack, line 74: `<version>${project.version}</version>`

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/impl/pipeline/graph/MergeStep.java
```diff
@@ -33,13 +35,14 @@
  * @author Alex Black
  */
 @AllArgsConstructor
-@EqualsAndHashCode(callSuper = true)
+@Data
```
AlexDBlack, line 38: Adding the `@Data` is ok but we still need the `@EqualsAndHashCode(callSuper = true)` here.
If you look at the build warnings, you'll find this:
```
Warning:(38, 1) java: Generating equals/hashCode implementation but without a call to superclass, even though this class does not extend java.lang.Object. If this is intentional, add '@EqualsAndHashCode(callSuper=false)' to your type.
```

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/impl/pipeline/graph/PipelineGraphStep.java
```diff
@@ -32,7 +33,6 @@
  * @author Alex Black
  */
 @Data
-@EqualsAndHashCode(callSuper = true)
```
AlexDBlack, line 35: Keep annotation

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/impl/pipeline/graph/SwitchOutput.java
```diff
@@ -34,7 +35,6 @@
  * @author Alex Black
  */
 @Data
-@EqualsAndHashCode(callSuper = true)
```
AlexDBlack, line 37: Keep

konduit-serving-pipeline/src/test/java/ai/konduit/serving/pipeline/impl/data/JsonCoverageTest.java
```diff
@@ -0,0 +1,251 @@
...
+    @Ignore
+    public void testInput() {
+        testConfigSerDe(new Input(new GraphBuilder().id("foo")));
+    }
+
+    // fails with https://gist.github.com/atuzhykov/fc87f4064ad00300da34403857d15d55
+    @Test
+    @Ignore
+    public void testPipelineProfilerTest() {
+        testConfigSerDe(new PipelineProfilerTest());
```
AlexDBlack, line 220: This is a test class, we don't need to test any of the test classes for JSON/YAML.
Let's add a method in BaseJsonCoverageTest like this:
```
public Set<Class<?>> ignores(){
    return Collections.emptySet();
}
```

For modules like this, we override that method in JsonCoverageTest and return the test classes.

Then in the for loop in `@Before` (i.e., in `for (Class<?> c : subTypes) {`) ignore if the class is in the set returned by ignores() method.

PipelineProfilerTest, FunctionStep, CallbackStep, are all like this


As for Input GraphStep (previous test) - that should never appear in the JSON (it's just for building) so let's ignore that one too.

konduit-serving-cli/pom.xml
```diff
@@ -105,6 +105,12 @@
             <artifactId>junit</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>ai.konduit.serving</groupId>
+            <artifactId>konduit-serving-common-tests</artifactId>
+            <version>0.1.0-SNAPSHOT</version>
```
AlexDBlack: For all of these: It's better practice to use `${project.version}` instead of the hardcoded version string like `0.1.0-SNAPSHOT`

konduit-serving-codegen/src/test/java/JsonCoverageTest.java
```diff
@@ -0,0 +1,46 @@
+/*
```
AlexDBlack, line 1: konduit-serving-codegen is an "old api" (to be deleted) module so this one can go

konduit-serving-common-tests/pom.xml
```diff
@@ -58,5 +57,24 @@
...
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+
+        </dependency>
+        <dependency>
+            <groupId>org.nd4j</groupId>
+            <artifactId>nd4j-common</artifactId>
+            <version>1.0.0-beta7</version>
```
AlexDBlack, line 77: `<version>${nd4j.version}</version>`

konduit-serving-data/konduit-serving-image/src/test/java/ai/konduit/serving/data/JsonCoverageTest.java
```diff
@@ -0,0 +1,275 @@
...
+public class JsonCoverageTest extends BaseJsonCoverageTest {
+
+    private ImageToNDArrayConfig c = new ImageToNDArrayConfig()
+            .height(256)
+            .width(256)
+            .channelLayout(NDChannelLayout.RGB)
+            .includeMinibatchDim(true)
+            .format(NDFormat.CHANNELS_LAST)
+            .dataType(NDArrayType.UINT8)
+            .normalization(new ImageNormalization(ImageNormalization.Type.SUBTRACT_MEAN));
```
AlexDBlack, line 63: For ImageNormalization - we should also test when meanRgb, stdRgb and maxValue are set

konduit-serving-data/konduit-serving-image/src/test/java/ai/konduit/serving/data/JsonCoverageTest.java
```diff
@@ -0,0 +1,275 @@
...
+    @Override
+    public Object fromYaml(Class<?> c, String yaml) {
+        return ObjectMappers.fromYaml(yaml, c);
+    }
+
+    @Test
+    public void testImageToNDArrayStep() {
+        testConfigSerDe(new ImageToNDArrayStep().config(new ImageToNDArrayConfig()
+                .height(128)
+                .width(128)));
```
AlexDBlack, line 84: More non-default values to test on ImageToNDArrayStep

konduit-serving-data/konduit-serving-image/src/test/java/ai/konduit/serving/data/JsonCoverageTest.java
```diff
@@ -0,0 +1,275 @@
...
+    }
+
+    @Test
+    public void testImageCropStep() {
+        testConfigSerDe(new ImageCropStep());
```
AlexDBlack, line 89: More non-default values on ImageCropStep to set for test

konduit-serving-models/konduit-serving-samediff/src/test/java/ai/konduit/serving/models/samediff/JsonCoverageTest.java
```diff
@@ -0,0 +1,34 @@
+package ai.konduit.serving.models.samediff;
```
AlexDBlack, line 1: Copyright header

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/impl/pipeline/graph/AnyStep.java
```diff
@@ -31,14 +33,14 @@
  *
  * @author Alex Black
  */
-@EqualsAndHashCode(callSuper = true)
+@Data
```
AlexDBlack, line 36: As already discussed in the previous PR review: We need to keep `@EqualsAndHashCode(callSuper = true)`. This same reason applies to every single one of these annotations that you've removed (not just the ones I've flagged)

konduit-serving-vertx-protocols/konduit-serving-grpc/src/test/java/ai/konduit/serving/vertx/protocols/grpc/test/JsonCoverageTest.java
```diff
@@ -0,0 +1,46 @@
+package ai.konduit.serving.vertx.protocols.grpc.test;/*
```
AlexDBlack, line 1: Copyright before package, not after.

konduit-serving-vertx-protocols/konduit-serving-http/src/test/java/ai/konduit/serving/vertx/protocols/http/test/JsonCoverageTest.java
```diff
@@ -0,0 +1,46 @@
+package ai.konduit.serving.vertx.protocols.http.test;/*
+ *  ******************************************************************************
```
AlexDBlack, line 2: Copyright before package

pom.xml
```diff
@@ -549,7 +549,7 @@
                 <groupId>junit</groupId>
                 <artifactId>junit</artifactId>
                 <version>${junit.version}</version>
-                <scope>test</scope>
+                <scope>compile</scope>
```
AlexDBlack, line 552: This has very big consequences. We cannot change this, it must stay test scope.

This means the junit testing library is now included in every single module at compile time, not just for tests.

PR #468, opened 2020-09-11
nd4j-tf step

atuzhykov:


agibsonccc (dismissed):
Minor nits and add tests?

konduit-serving-models/pom.xml
```diff
@@ -34,6 +34,7 @@
         <module>konduit-serving-tensorflow</module>
         <module>konduit-serving-samediff</module>
         <module>konduit-serving-onnx</module>
+        <module>konduit-servong-nd4j-tensorflow</module>
```
agibsonccc, line 37: typoe

agibsonccc (commented):
@atuzhykov you're missing the extra files needed for module definitions as well (this is for java 9+) could you make sure to copy the same setup for modules under src/main/resources (META-INF/services) and also add the associated *Module.java? Thanks!

agibsonccc:
@atuzhykov  you'll want to look in to why the graph runner init method isn't being called:
https://github.com/KonduitAI/deeplearning4j/blob/master/nd4j/nd4j-tensorflow/src/main/java/org/nd4j/tensorflow/conversion/graphrunner/GraphRunnerServiceProvider.java#L18
Make sure to call init when the step is created.

ShamsUlAzeem (approved):
LGTM +1

PR #469, opened 2020-09-16
Python Paths resolution ADR

ShamsUlAzeem:
Proposal for a workflow for dynamic python paths resolution.


ADRs/0011-Python-Path-Resolution.md
```diff
@@ -0,0 +1,306 @@
...
+
+## Status
+PROPOSED
+
+Proposed by: Shams Ul Azeem (15/09/2020)
+
+Discussed with: Adam Gibson
+
+## Context
+Konduit Serving has the ability to run a python script using a PythonStep. There are a bunch of options that need to be setup for the PythonStep which includes, names and data types for inputs & outputs, script locations, python library paths. All of these options remain static from system to system apart from the python library paths variable. Since, every system has a different set of python installations using either conda, virtual environments or regular python installs, it becomes a hassle to manage a PythonStep configuration that's widely usable across different systems. The point of creating this ADR is to specify a simple enough workflow that can take care of the python library paths resolution for us through both using the konduit-serving cli **profiles** and simple configs.
```
agibsonccc, line 11: Maybe add a bit more about the auto discovery? I'd also add a registration command like konduit register path/to/python/binary kind of like what you do in pycharm.

ADRs/0011-Python-Path-Resolution.md
```diff
@@ -0,0 +1,306 @@
...
+                               
+### Defaults and Priorities
+The defaults for python paths are to be expected in the following way:
+
+![Python Paths Defaults](../markdown_images/python-installation-paths-defaults.png)
+
+### Usage with Profiles
+The profiles should be extended to incorporate the python paths settings for running with the `serve` command. The same items names that are used for the PythonStep config are going to be used here. For example: 
+
+```shell script
```
agibsonccc, line 135: Specify a bit  about how this will work? Also if we're pointing at a conda environment one thing we might want to do is specify a conda environment to use?

agibsonccc (changes requested):
Nits

ADRs/0011-Python-Path-Resolution.md
ShamsUlAzeem: Ah yes! The I'll add the details about the auto discovery. Basically it's going to find the python and conda executables in the PATH environment variable and then assign their respective IDs. 

Also, the registration command is a good idea. It will register and keep it in the python paths even if it's not inside the **PATH** env variable.

ADRs/0011-Python-Path-Resolution.md
ShamsUlAzeem: Yup the shell command will take care of that. You'll have  an option to specify a conda environment name as well when you put `type == conda` in the config.

ShamsUlAzeem: so this command has that option here: https://github.com/KonduitAI/konduit-serving/pull/469/files#diff-87058a0e57d0b969b3864dc5a0e0cbe8R136 under `--conda-env` or `-ce` flag

ADRs/0011-Python-Path-Resolution.md
```diff
@@ -0,0 +1,322 @@
...
+konduit profile create python_37_env --python-type=conda --python-location=1 --conda-env=py37 <...other_options...>
+                                        # OR
+konduit profile create python_37_env -pt=conda -pl=1 -ce=py37 <...other_options...>
+```
```
agibsonccc, line 139: Another note might be virtual envs as well. I'm not sure how hard it is, but specifying whether we create a separate virtual env could also be good. Same for running conda. I wonder if we could bootstrap a conda environment from   konduit as well. As long as we have access to a python binary we're good to go.

ADRs/0011-Python-Path-Resolution.md
```diff
@@ -0,0 +1,322 @@
...
+    -------------------------------------------------------------------------------------------------
+-----------------------------------------------------------------------------------------------------
+```
+
+Based on the above info (gained from the command: `konduit pythonpaths`), it's much easier to specify python paths in configuration files (using `type`, `location`, `environment` properties):
+
+```json
+{
+  "@type" : "PYTHON",
+  "pythonConfig" : {
```
agibsonccc, line 82: Could you show a code example as well? It'd be good to see something like pythonPathProfileId(..) as well.

Specify basic validation as well like: "couldn't find path.." etc

agibsonccc (changes requested):

ADRs/0011-Python-Path-Resolution.md
ShamsUlAzeem: Hmm. Virtualenv would more or less have the same workflow so shouldn't be hard to add it here. We basically have a lot of options to do here but most of it is running conda/python/virtualenv commands. There's no problem using the basic few commands for listing packages or versions but using dependencies resolution for python would take a bit more effort here. It's fine for cases where there are no clashes between deps but resolving clashes is gonna be a mess if we try to do it directly from konduit.

ADRs/0011-Python-Path-Resolution.md
ShamsUlAzeem: Sure. I'm calling it location for now for the custom python type where `location=E:\python37\python.exe` could make more sense

ADRs/0011-Python-Path-Resolution.md
ShamsUlAzeem: I'll keep bootstrapping new python environments (whether `virtualenv` or `conda`) for a separate PR. For this ADR, I'll just work on specifying pre-installed/created python installs/environments.

ADRs/0011-Python-Path-Resolution.md
ShamsUlAzeem: I'll add an example end-to-end workflow

ShamsUlAzeem: I've changed `location` to just `path` in the examples. They're not exactly end-to-end but they give you the whole workflow idea pretty well.

treo:
I've got a few questions that came up as I was reading the ADR:
* Why are we using the system python? Can we not ship our own python?
* The way this is supposed to work, binds it pretty hard to the defining system. What if id 1 is a different kind of python on a different system?
* There is a bit of magic in the defaults, all "magic" should be explicitly logging what it is doing, so users can actually know what is happening - or why something doesn't work anymore after they've moved their jar file to a different system
* How does this work with deployment type "jar"?
* Static Setting / different python paths for different steps: How does this work for the "create on one machine, serve on a different one" use case?

Overall, I'm not sure the direction of the PythonStep is what I would expect given the options we have. the cpython bytedeco package can provide us with a known python executable in a known place and it does ship with both pip and venv. On that grounds I'd expect the python step to be able to be aware of a `requirements.txt` file, so it should also be able to set up all necessary libraries in an environment for that python step.

ShamsUlAzeem:
@treo 

Thanks Paul, appreciate the review. I like the suggestion for `venv` for cpython for bootstrapping new environments from a requirements.txt file. I'll answer the queries one by one:

- **Why are we using the system python? Can we not ship our own python?**
We don't have to necessarily use the system available python installation. We have that as an option so it can work without having to add anything extra on top of the konduit installation and a working configuration. If we have a separate python installation script or a zip we can definitely extract that to a particular location and specify the its path with `type=custom` & `path=<install_location>` and it should be able to work.

- **The way this is supposed to work, binds it pretty hard to the defining system. What if id 1 is a different kind of python on a different system?**
So, like mentioned in the previous answer, it doesn't have to use the system python. It can use a separate folder where we have the python binaries present in the system. I would consider it as a utility to manage python installations on various locations using different tools. The way it's going to work with different systems is by using profiles that can be setup separately for each system. The only thing that will need to change here is the `serve` command with the right profile. For example: 

```
konduit serve -c config.yaml -id python-server --profile python_profile
```

That way even if the config is transferred from another system the profile will make sure the right python is used even if the configuration is different settings. The profile settings would override what's present in the python step config.

- **There is a bit of magic in the defaults, all "magic" should be explicitly logging what it is doing, so users can actually know what is happening - or why something doesn't work anymore after they've moved their jar file to a different system.**
Yup, every activity is going to be logged while the python step is being initialised so that it's prominent if the right behaviour is being followed. I should demonstrate the workflow in the ADR as well. 

- **How does this work with deployment type "jar"?**
Not sure if I understand this question correctly but if you mean how it would work with a konduit-serving uber-jar then the uber-jar would have all the CLI features so all the commands related to python paths (`pythonpaths`, `pythonpaths add`) would be available through that as well. So, we can simply do something like: `java -jar konduit-serving.jar pythonpaths` and see all the available options there. Also, profiles would be available as well. 

- **Static Setting / different python paths for different steps: How does this work for the "create on one machine, serve on a different one" use case?**
I thought of this for a use case where a user might want to run some python steps on a CPU and some on GPU. For example, I can have a python step that just wants to use tensorflow and another that might wanna use tensorflow-gpu. This would be largely machine specific. So what profiles would do is that they will override the config settings for python paths. Keeping `python-path-resolution: static` would stop it from doing that. The default value of that setting would be `python-path-resolution: dynamic` where profiles would affect the step configuration. 

---

The cpython venv idea is a good option to have. We can incorporate that in current workflow with adding a few more types like:

```
{
    type: bootstrap
    requirements: <path-of-requirements.txt>
    path: <path-to-new-venv>
    .
    .
    .
}

However, I was thinking of keeping all the installation and validation details this type of python specification as a separate ADR since there would be a lot of edge cases where the installation fails and how we can deal with that.

**Comment by treo** [comment]
> I like the suggestion for venv for cpython for bootstrapping new environments from a requirements.txt file.
> If we have a separate python installation script or a zip we can definitely extract that to a particular location and specify its path with type=custom & path=<install_location> and it should be able to work.

I'm not quite sure my point about cpython came across correctly. What I specifically mean is that the following exists:
```xml
<dependency>
  <groupId>org.bytedeco</groupId>
  <artifactId>cpython</artifactId>
  <version>3.7.9-1.5.4</version>
</dependency>
```

The platform specific jar files for it contain a full python installation, along with the standard library. 

The PythonStep is already depending on `org.bytedeco:numpy-platform` and that in turn does depend on the cpython package that I'm talking about.

Now that I've taken a look at the code, I see that the PythonStepRunner is already using PythonExecutioner from datavec, which in turn actually *does* use that package through the Java Interface, and that in turn means that it already uses the internal interpreter. 

I think that the proposal is somewhat misguided given this knowledge. I fully expect things to crash in hard to debug ways, if we are using the bundled python executable with a different set of libraries that at best expect a slightly different standard lib and at worst where compiled for a different python version.

ShamsUlAzeem:
So, based on our call. Javacpp cpython distro has most of the startup packages (venv, pip) that we need to bootstrap compatible environments. In case of system installed packages, we might lead into some cases where the compiled packages or different python versions are just not compatible with the cpython that gets shipped with javacpp.

We can do some validations here and there but the problem isn't fully solved since the interpreter being used is going to be the same in all cases (that comes with javacpp cpython). So, included venv & pip is the safest way to go.

For no/bad internet case. We can have a simple python path (executable) option that can just take a python executable location and run python step with the libraries included in that package. For packages clashes we do have options for appending or excluding libraries path with javacpp packages.

I'll create another ADR that mentions all of these points and then we can decide which direction we should go.

ADRs/0011-Python-Path-Resolution.md
```diff
@@ -0,0 +1,631 @@
...
+.
+.
+```
+
+Now we want to specify the python install with `id = 4` into the PythonStep configuration in Java. The code would look similar to: 
+
+```java
+PythonConfig pythonConfig = PythonConfig.builder()
+        .pythonType(PythonConfig.PythonType.Python)
+        .pythonPath("4")
```
agibsonccc, line 329: We probably should give the user the option to name the pipeline. IDs will be hard to remember.

ADRs/0011-Python-Path-Resolution.md
```diff
@@ -0,0 +1,631 @@
...
+.
+.
+```
+
+Now we want to specify the conda install with `id = 1` and the conda environment `py37` into the PythonStep configuration in Java. The code would look similar to: 
+
+```java
+PythonConfig pythonConfig = PythonConfig.builder()
+        .pythonType(PythonConfig.PythonType.Conda)
+        .pythonPath("1")
```
agibsonccc, line 371: We might want to do an object oriented/inherited approach for python paths/types to allow us to add specific variables for different environments. We risk the same kind of sprawl we used to have before otherwise. We will probably need to add a lot of specific variables for conda.

agibsonccc (changes requested):

agibsonccc:
@ShamsUlAzeem in testing this, one thing we need to make sure of is having a feedback loop for adding python paths. Right now it doesn't print anything. Something like:
Successfully added xyz python path to config file abc.json would go a long way to helping people understand what the cli is doing.

agibsonccc:
@ShamsUlAzeem more feedback: How do we handle the default conda environment? Do we allow the base environment? We should have explicit behavior for this whether that's a default value or throwing an exception if the base environment is detected.

PR #470, opened 2020-09-23
ADR 0011 Implementation - Python Paths Resolution

ShamsUlAzeem:
Implementation of ADR 0011 for resolving python paths for the PythonStep configuration.


konduit-serving-cli/src/main/java/ai/konduit/serving/cli/launcher/command/build/extension/PythonPathsCommand.java
```diff
@@ -0,0 +1,243 @@
...
+                String.format(
+                                "\t--------------------------ENVIRONMENTS-------------------------%n" +
+                                "\t%s" +
+                                "\t---------------------------------------------------------------%n",
+                        String.join(System.lineSeparator() + "\t", formattedCondaEnvironments)
+                )
+        );
+    }
+
+    private static void listVenvInstallations() {
```
agibsonccc, line 238: Throw an exception here so we know it's not supported yet.

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/api/process/ProcessUtils.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.stream.Collectors;
+
+@Slf4j
+public class ProcessUtils {
+
+    private static String OS = System.getProperty("os.name").toLowerCase();
+
+    public static boolean isWindows() {
+        return OS.contains("win");
+    }
```
agibsonccc, line 39: I would recommend zt-exec here instead of doing things like this.

agibsonccc (changes requested):
Minor nits, pretty good so far!

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/api/process/ProcessUtils.java
ShamsUlAzeem: Ah yes! Thanks for suggesting this. It really makes it easy to work with processes. I was trying to keep the `konduit-serving-pipeline` module clean from extra dependencies since I've already broken a rule by adding the ProcessUtils logic in there.

konduit-serving-cli/src/main/java/ai/konduit/serving/cli/launcher/command/build/extension/PythonPathsCommand.java
ShamsUlAzeem: Great idea

ShamsUlAzeem: Adding them won't be much work so I'll complete them in this PR as well

konduit-serving-api/src/main/java/ai/konduit/serving/model/PythonConfig.java
```diff
@@ -65,7 +80,140 @@
     @Builder.Default
     private String jobSuffix = "konduit_job";
 
+    public String resolvePythonLibrariesPath() {
+        if(pythonConfigType == null) {
+            log.info("Python config type not specified...");
+            List<CondaDetails> condaInstalls = PythonPathUtils.findCondaInstallations();
```
agibsonccc, line 86: I would only do this if a conda environment type is specified and no environment name is specified. Force the user to be explicit about what they're using.

konduit-serving-cli/pom.xml
```diff
@@ -38,7 +38,25 @@
             </dependency>
         </dependencies>
     </dependencyManagement>
+
     <dependencies>
+        <dependency>
```
agibsonccc, line 43: Is this needed? All the javacpp deps should already be a transitive dependency.

agibsonccc (changes requested):

konduit-serving-cli/pom.xml
ShamsUlAzeem: They are but those deps are in provided scope so they won't be included in the cli jar or you can say that it can't be guaranteed that it's gonna be present

konduit-serving-api/src/main/java/ai/konduit/serving/model/PythonConfig.java
ShamsUlAzeem: All of the activity is going to be logged to I don't see a big problem here. I was trying to have something where a person has some variables defined and can use the python config without explicitly specifying python paths.

ShamsUlAzeem: Unless he has nothing installed in his environment at all so it's not gonna use javacpp by default and instead prompt the user to use the locally installed python at a custom place.

agibsonccc: The issue is it assumes conda. Javacpp would be reasonable since it's bundled.

PR #474, opened 2020-10-27
Clean up deprecated files

agibsonccc:
Removes deprecated files.


ShamsUlAzeem (approved):
Good stuff! About time these are going out of the codebase

PR #482, opened 2020-12-25
Adding default metrics and image + array resolving in predict endpoint

ShamsUlAzeem:


konduit-serving-cli/pom.xml
```diff
@@ -61,6 +61,11 @@
             <groupId>com.github.oshi</groupId>
             <artifactId>oshi-core</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.tika</groupId>
+            <artifactId>tika-core</artifactId>
+            <version>1.25</version>
```
agibsonccc, line 67: Mind adding a ${tika.version}?

konduit-serving-data/konduit-serving-nd4j/src/main/java/ai/konduit/serving/data/nd4j/format/ND4JNDArrayFactory.java
```diff
@@ -51,6 +57,14 @@ public NDArray create(Object o) {
         INDArray a;
         if(o instanceof INDArray){
             a = (INDArray)o;
+        } else if (o instanceof ListValue) {
+            List<Long> shape = new ArrayList<>();
+            List<Float> data = new ArrayList<>();
+
+            ListValue listValue = (ListValue) o;
```
agibsonccc, line 64: Could you explain the point of the ListValue?

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/impl/serde/DataJsonSerializer.java
```diff
@@ -177,33 +178,50 @@ private void writeImage(JsonGenerator jg, Image i) throws IOException {
     }
 
     private void writeNDArray(JsonGenerator jg, NDArray n) throws IOException {
-        jg.writeStartObject();
+        if(n.rank() > 0 && n.rank() < 5) {
+            switch (n.rank()) {
```
agibsonccc, line 182: Maybe you could have reused the NDArrayStrings class in dl4j? It outputs valid json already.

konduit-serving-vertx-protocols/konduit-serving-http/pom.xml
```diff
@@ -63,5 +63,10 @@
             <version>0.1.0-SNAPSHOT</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>ai.konduit.serving</groupId>
```
agibsonccc, line 67: Mind changing all 0.1.0-SNAPSHOT to ${project.version}?

konduit-serving-vertx-protocols/konduit-serving-http/src/main/java/ai/konduit/serving/vertx/protocols/http/verticle/InferenceVerticleHttp.java
```diff
@@ -149,7 +153,7 @@ private HttpServerOptions createOptions(int port, boolean useSsl, String sslKeyP
                 .setCompressionSupported(true)
                 .setTcpKeepAlive(true)
                 .setTcpNoDelay(true)
-                .setAlpnVersions(Arrays.asList(HttpVersion.HTTP_1_0,HttpVersion.HTTP_1_1))
+                .setAlpnVersions(Arrays.asList(HttpVersion.HTTP_1_0, HttpVersion.HTTP_1_1))
```
agibsonccc, line 156: A lot of browsers are going to force http2 soon. This might not be such a good idea.

agibsonccc (changes requested):

konduit-serving-data/konduit-serving-nd4j/src/main/java/ai/konduit/serving/data/nd4j/format/ND4JNDArrayFactory.java
ShamsUlAzeem: Similar to the difference between list and numpy.ndarrays in python. Other than that I don't think there's much use here

ShamsUlAzeem: I had to do this so that the json arrays passed as inputs have a way to be converted into INDArrays since it's inconvenient to find a object representation of INDArray and pass that in http requests

konduit-serving-pipeline/src/main/java/ai/konduit/serving/pipeline/impl/serde/DataJsonSerializer.java
ShamsUlAzeem: I can't use it here since it's part of the `nd4j-api` library and by design it's not supposed to be a part of `konduit-serving-pipeline`

konduit-serving-data/konduit-serving-nd4j/src/main/java/ai/konduit/serving/data/nd4j/format/ND4JNDArrayFactory.java
ShamsUlAzeem: Other than that ListValues are useful for sending nested java as well

ShamsUlAzeem: Also useful for sending in other data types like list of strings or list of objects

agibsonccc (approved):

PR #511, opened 2022-02-17
Changes to sd.output since samediff handles sessions internally 

agibsonccc:
Change samediff runner to use Change to sd.output(..) instead of inferencesession directly due to performance reasons.


agibsonccc:
@quickwritereader

